PR #16, opened 2014-10-22
Add sale_condition_template

guewen:
I moved this module from sale-workflow.  Displaying additional condition texts on sales reports is clearly a reporting module.

This module has been migrated from sale_condition_text, here is the original PR for the migration before I moved the module: https://github.com/OCA/sale-workflow/pull/50

This is still a WIP module because the report does not show the fields at the moment.

Depends on :
- [x] https://github.com/OCA/account-invoice-reporting/pull/14
- [x] https://github.com/OCA/account-invoice-reporting/pull/15


guewen:
Don't be scared by some old commits, I kept the original and full history from the `sale-workflow` repository.

sale_condition_template/sale_order.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def set_note1(self, cr, uid, so_id, cond_id, partner_id):
+        return self.set_condition(cr, uid, cond_id, 'note1', partner_id)
+
+    def set_note2(self, cr, uid, so_id, cond_id, partner_id):
+        return self.set_condition(cr, uid, cond_id, 'note2', partner_id)
+
+    def action_invoice_create(self, cr, uid, ids,
+                              grouped=False,
+                              states=['confirmed', 'done', 'exception'],
```
guewen: Pylint says with reason:

> sale_condition_template/sale_order.py:52: [W0102(dangerous-default-value), SaleOrder.action_invoice_create] Dangerous default value ['confirmed', 'done', 'exception'] as argument

How should we handle that as the same default value is defined in the `super` method in the keyword argument? If we send `None`, the `super` method will have `None` instead of `['confirmed', 'done', 'exception']`.
Ideally it should be fixed in the core by setting `None` by default and with this code at the start of the method:

```
if states is None:
    states = ['confirmed', 'done', 'exception']
```

Now, in this module, I see 2 options:

A. None as sentinel and default value

```
def action_invoice_create(self, cr, uid, ids,
                          grouped=False,
                          states=None,
                          date_invoice=False, context=None):
    if states is None:
        states = ['confirmed', 'done', 'exception']
    _super = super(SaleOrder, self)
    inv_id = _super.action_invoice_create(cr, uid, ids,
                                          grouped=grouped,
                                          states=states,
                                          date_invoice=date_invoice,
                                          context=context)
```

B. Do not propagate the argument if not provided

```
def action_invoice_create(self, cr, uid, ids,
                          grouped=False,
                          states=None,
                          date_invoice=False, context=None):
    _super = super(SaleOrder, self)
    if states is None:
        inv_id = _super.action_invoice_create(cr, uid, ids,
                                              grouped=grouped,
                                              date_invoice=date_invoice,
                                              context=context)
    else:
        inv_id = _super.action_invoice_create(cr, uid, ids,
                                              grouped=grouped,
                                              states=states,
                                              date_invoice=date_invoice,
                                              context=context)
```

The option B. seems farfetched but it seems better to me because it does not duplicate the default value `['confirmed', 'done', 'exception']`.

guewen:
It still misses the report part.

sale_condition_template/sale_order.py
yvaucher: Maybe we could add a dict for kwargs to use the correct arguments for option B

guewen: Adopted the @yvaucher's proposal, thanks

guewen:
I just added the fields on the report.

sale_condition_template/README.rst
```diff
@@ -0,0 +1,28 @@
+Sale Conditions
+===============
+
+Adds template condition texts on sales orders and invoices.
```
yvaucher, line 4: Description to change, this modules doesn't add template on invoices here is a proposal:

```
Complete module `Invoice conditions` by adding template condition texts
on sales orders.

Templates are categorized by their position on the document.

Two positions are available:

- before sale order lines
- after sale order lines

Texts are copied on the invoice when you will create invoice from sale order.
```

yvaucher:
apart from description :+1:

jgrandguillaume:
Hi,

Thanks for the work ! What I saw:
- We need a way to manage (create and delete) the sales conditions. Can you add a menu in sales configuration ?
- In SO form, it become difficult to see everything. Can you put all top and bottom condition in a new tab ?

Otherwise, it's a great job.

Regards,

Joël

yvaucher:
@jgrandguillaume I wanted to make a "what you see is what you get".

Previously we had all notes at the same place as the additional informations. Plus templates m2o on an other tab.

I agree that with big condition your lines will step on the bottom. However those conditions shouldn't take more place than we have on the invoice page.
In this new tab we may need to add some more informative text to not let the user lost.

An other way I could think of is to collapse the condition text and uncollapse when editing it to be more readable but not sure we can do this in an easy way.

guewen:
Thanks for your review @jgrandguillaume,
The menu is already there: https://github.com/OCA/sale-reporting/pull/16/files#diff-253f4922b5fbca2cd12410ab96c5cd72R4

I agree that the text fields take too much place and we should find a better way to display them.

Your review on this one would be nice too because it is a pre-requisite: https://github.com/OCA/account-invoice-reporting/pull/15

gurneyalex:
:+1: waiting for travis to say "green"!

PR #18, opened 2014-11-26
Rename sale_condition_template and improve its UI

guewen:
Move the condition fields in a new page

They take too much place on the 'Order Lines' page. Also I improved the
general layout of the fields.

Rename 'Condition' to 'Comment' because the texts are not 'conditions'

I will do the same changes to invoice_condition_template https://github.com/OCA/account-invoice-reporting/pull/17


pedrobaeza:
If you have changed the term from 'conditions' to 'comments', module should be also renamed to sale_comment_template?

guewen:
@pedrobaeza renaming a module can break installations

guewen:
BTW I changed the 'name' in `__openerp__.py` so it is displayed as Sale Comments in Local modules / Apps.

pedrobaeza:
The life of the module is very short, so I don't think it's going to impact, but as you want. If not renamed, we should be aware for making the rename on v9.

guewen:
> The life of the module is very short, so I don't think it's going to impact, but as you want. If not renamed, we should be aware for making the rename on v9.

You are right. I thought that it was the same name in v7, but it isn't. I will propose the change.

(but I need to do the same thing on the base and invoice modules)

guewen:
I renamed the module, as well as base_condition_template and account_condition_template in https://github.com/OCA/account-invoice-reporting/pull/17

guewen:
Build fails now because the dependency module has been renamed in https://github.com/OCA/account-invoice-reporting/pull/17

pedrobaeza:
:+1:

pedrobaeza:
Please rename also pot file.

guewen:
Pot file renamed. And I restarted the build now that OCA/account-invoice-reporting#17 is merged

pedrobaeza:
This can be fast-tracked also.

yvaucher:
:+1:

PR #22, opened 2015-01-21
Use the _prepare method instead of write

guewen:
It also fixes an assert when action_invoice_create is called
on several ids

```
assert len(ids) == 1, "1 ID expected, got: %s" % (ids, )
AssertionError: 1 ID expected, got: [39, 38]
```


pedrobaeza:
Good option :+1:

yvaucher:
:+1:

gdgellatly:
:+1:

PR #27, opened 2015-07-20
[ADD] Module for sale reporting in company currency

mistotebe:


pedrobaeza:
This should be in sale-workflow repo, but why overwrite current standard compute methods instead of adding fields? It's very dangerous to change this behaviour.

mistotebe:
Isn't that exactly what I do, add fields and hooking into the current methods? If you mean hooking into _amount_all, I guess I could experiment whether making them fully new style with proper dependencies would work.

My thinking is that there is no workfllow involved, only making the data available for reporting, hence proposing it here.

pedrobaeza:
Call it then `sale_report_...`. Fields in sale.order are not being displayed in any place.

mistotebe:
I have renamed the package and the sales team tickers now use the new fields.

mistotebe:
And switched to using new style function fields for them as well.

pedrobaeza:
Thanks for the changes.

:+1:

sale_report_company_currency/README.rst
```diff
@@ -0,0 +1,60 @@
...
+* Users having access to multiple companies with differing currencies might
+  still get a useless cross-currency sum in some reports.
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/sale-reporting/issues>`_.
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
+`here <https://github.com/OCA/sale-reporting/issues/new?body=module:%20sale_report_company_currency%0Aversion:%20{version}%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
```
eLBati: Please replace `{version}`

mistotebe:
@eLBati: thanks, have updated the URL now.

sale_report_company_currency/README.rst
```diff
@@ -0,0 +1,56 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :alt: License: AGPL-3
+
+Sale reporting in company currency
+==================================
+
+This module records sale order proces in company currency along the pricelist
```
gurneyalex: s/proces/prices/ (?)
s/along the/along with the/

pedrobaeza:
This has been fixed meanwhile in upstream, so I close.

PR #34, opened 2016-08-25
Update sale_comment_template to odoo 9

leNeo:


leNeo:
Migration to version 9.0

PR #36, opened 2017-01-11
[ADD] sale_reporting_weight

pedrobaeza:
Weights in the sales analysis view
==================================

This module adds the measure "Weight" in the sales analysis view. This is caught from 2 possible sources:

* If the UoM of the product is one of the category "Weight", the value is taken from the ordered quantity.
* If the UoM of the product is another, then the weight is taken from the weight field of the product multiply by the ordered quantity.

Usage
=====

* Go to *Sales > Reports > Sales*.
* Add the "Weight" measure from your "Measures" dropdown in your analysis.

Known issues / Roadmap
======================

* The weight quantity is expressed in the unit of measure of the product, so if you have several weight UoMs across your products, the global sum won't make sense.


pedrobaeza:
cc @Tecnativa

PR #45, opened 2018-01-23
[ADD] sale_comment_template: add migration scripts

cubells:
cc @Tecnativa

### DIFF
--- a/sale_comment_template/migrations/8.0.1.0.0/post-migration.py
+++ b/sale_comment_template/migrations/8.0.1.0.0/post-migration.py
@@ -1,0 +1,32 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Tecnativa - Vicent Cubells
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+
+
+def migrate_comment_template(cr):
+    cr.execute(
+        """
+        UPDATE sale_order so
+        SET comment_template1_id = bct.condition_template_id,
+            note1 = bct.text
+        FROM base_comment_template bct
+        WHERE bct.id = so.condition_template_id
+        AND bct.position = 'before_lines'
+        """)
+    cr.execute(
+        """
+        UPDATE sale_order so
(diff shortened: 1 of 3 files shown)


sale_comment_template/migrations/8.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Tecnativa - Vicent Cubells
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from openupgradelib import openupgrade
+from openerp.modules.registry import RegistryManager
```
pedrobaeza: Why importing this if it's not going to be used?

sale_comment_template/migrations/8.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,25 @@
...
+
+def migrate_comment_template(cr, registry, template):
+    cr.execute(
+        """
+        UPDATE sale_order so
+        SET comment_template1_id = bct.{template_id},
+            note1 = bct.text
+        FROM base_comment_template bct
+        WHERE bct.{comment_id} = so.id
+        """.format(template_id=template))
```
pedrobaeza: Where is `comment_id`?

sale_comment_template/migrations/8.0.1.0.0/pre-migration.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+column_renames = {
+    'sale_order': [('condition_template_id', None)],
+}
+
+
+@openupgrade.migrate()
+def migrate(cr, version):
+    openupgrade.rename_columns(cr, column_renames)
```
pedrobaeza: Do we really need to rename column? Remember that SQL columns are not dropped, so the only reason for renaming is because there's another column with the same name with other function or other values.

pedrobaeza (changes requested):

yajo (approved):

pedrobaeza (approved):

PR #50, opened 2018-07-10
[ADD] sale_report_delivered_subtotal: New Module

chienandalu:
Adds subtotal price based on the delivered quantities field to the
*Sales Report*.

cc @Tecnativa

### DIFF
--- a/sale_report_delivered_subtotal/__manifest__.py
+++ b/sale_report_delivered_subtotal/__manifest__.py
@@ -1,0 +1,15 @@
+# Copyright 2018 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+{
+    "name": "Sale Report Delivered subtotal",
+    "version": "11.0.1.0.0",
+    'author': 'Tecnativa,'
+              'Odoo Community Association (OCA)',
+    'website': 'https://github.com/OCA/sale-reporting',
+    "category": "Sales",
+    "license": "AGPL-3",
+    "depends": [
+        "sale",
+    ],
+    "installable": True,
+}
--- a/sale_report_delivered_subtotal/readme/USAGE.rst
+++ b/sale_report_delivered_subtotal/readme/USAGE.rst
@@ -1,0 +1,2 @@
+Go to *Sales > Reporting > Sales* and you'll find the new *Total Price
+Delivered* measure.
--- a/sale_report_delivered_subtotal/reports/sale_report.py
+++ b/sale_report_delivered_subtotal/reports/sale_report.py
@@ -1,0 +1,21 @@
+# Copyright 2018 Tecnativa - Sergio Teruel
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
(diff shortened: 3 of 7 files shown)


sale_report_delivered_subtotal/reports/sale_report.py
```diff
@@ -0,0 +1,20 @@
...
+
+    price_subtotal_delivered = fields.Float(
+        string='Total Price Delivered',
+        readonly=True,
+    )
+
+    def _select(self):
+        select_str = super(SaleReport, self)._select()
+        select_str += """,
+            sum(qty_delivered * l.price_unit) as price_subtotal_delivered
```
pedrobaeza: In case that taxes are included in the price, this won't be correct, so the correct tentative formula is:
```
sum(l.price_subtotal / l.product_uom_qty * l.qty_delivered) AS price_subtotal_delivered
```

pedrobaeza (changes requested):

pedrobaeza (approved):

chienandalu:
ping @sergio-teruel

sale_report_delivered_subtotal/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+{
+    "name": "Sale Report Delivered subtotal",
+    "version": "11.0.1.0.0",
+    'author': 'Tecnativa,'
+              'Odoo Community Association (OCA)',
+    'website': 'https://github.com/OCA/sale-reporting',
+    "category": "Sales",
+    "license": "AGPL-3",
+    "depends": [
+        "sale",
```
rafaelbn, line 12: IMHO this modules should depends in sale_management. Without sale_management you cannot use this module...

sale_report_delivered_subtotal/readme/USAGE.rst
```diff
@@ -0,0 +1,2 @@
+Go to *Sales > Reporting > Sales* and you'll find the new *Total Price
+Delivered* measure.
```
rafaelbn, line 2: @chienandalu @sergio-teruel the field is OK but the screenshot...



It looks like Odoo has a bug

rafaelbn (approved):
Functionally tested :+1: 

Thanks

sale_report_delivered_subtotal/__manifest__.py
chienandalu: I'm not completely sure about it. Isn't the rule to pull the minimum dependencies?

sale_report_delivered_subtotal/readme/USAGE.rst
chienandalu: I'd say we can't do much :confused:

sale_report_delivered_subtotal/__manifest__.py
pedrobaeza: Yes, you shouldn't put not technically needed dependencies.

pedrobaeza:
Please squash before merging

chienandalu:
@pedrobaeza Done

PR #42, opened 2017-08-23
[10.0] [ADD] sale_order_report_with_note

nikul-serpentcs:
Sale Order Report With Note
=============================

* This module replaces the WebKit Quotation / Sale Order report by  a qweb report and adds header and footer notes on the sale order  and quotation

PR #51, opened 2018-07-19
[10.0][ADD] sale_order_report_with_note

MiquelRForgeFlow:
Superseeds https://github.com/OCA/sale-reporting/pull/42.


MiquelRForgeFlow:
@aheficent could you check and review?

AaronHForgeFlow (changes requested):
Both, headers and footers are duplicated in the report. And the view is a bit ugly:
image

AaronHForgeFlow:
Trying to create footer text on runbot and got the following:

>   File "/home/odoo/build/OCA/sale-reporting/sale_order_report_with_note/models/sale.py", line 88, in set_condition_header
    self.text_condition2.text = self.note2
  File "/.repo_requirements/odoo/odoo/fields.py", line 930, in __set__
    record.ensure_one()
  File "/.repo_requirements/odoo/odoo/models.py", line 4844, in ensure_one
    raise ValueError("Expected singleton: %s" % self)
ValueError: Expected singleton: sale.condition_text()

Also cannot click on create and edit in header and footer conditions

MiquelRForgeFlow:
Fixed all errors. Now it works fine!

sale_order_report_with_note/__manifest__.py
```diff
@@ -0,0 +1,27 @@
...
+{
+    'name': 'Sales Order Report With Note',
+    'version': '10.0.1.0.0',
+    'category': 'Reports/Qweb',
+    'license': 'AGPL-3',
+    "author": "Camptocamp SA,"
+              "Eficent, "
+              "Serpent CS,"
+              "Odoo Community Association (OCA)",
+    'website': "https://odoo-community.org/",
```
bodedra, line 17: shouldn't be https://github.com/OCA/sale-reporting ?

MiquelRForgeFlow: yes

MiquelRForgeFlow:
But wasted work. Unfortunately, this PR should stall and not be merged. The module [sale_comment_template](https://github.com/OCA/sale-reporting/tree/10.0/sale_comment_template) already has the same features.

MiquelRForgeFlow:
Features of this module moved to https://github.com/OCA/sale-reporting/pull/52.

AaronHForgeFlow:
I think this can be closed @mreficent

MiquelRForgeFlow:
Right

PR #53, opened 2018-07-31
[11.0][MIG] sale_comment_template

xavierjimenez:
Migration of the module **sale_comment_template** to version 11.0.

Depends on:

- [x] https://github.com/OCA/account-invoice-reporting/pull/81 invoice_comment_template


sale_comment_template/README.rst
```diff
@@ -0,0 +1,64 @@
...
+- above sale order lines
+- below sale order lines
+
+Texts are copied on the invoice when you will create invoice from sale order.
+
+Installation
+============
+
+To install this module, you need to have the module `Invoice Comments
+<https://github.com/OCA/account-invoice-reporting/tree/10.0/invoice_comment_template>`_.
```
bodedra: change Odoo version from 10.0 to 11.0

sale_comment_template/README.rst
```diff
@@ -0,0 +1,64 @@
...
+
+Usage
+=====
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/148/10.0
```
bodedra: Same as above.

HaraldPanten:
Hi @xavierjimenez ,

As a suggestion, I think it would be useful to add a menu item in Sales --> Configuration to access the templates.

If not, you can only access to the "Document Comments" from the Invoicing --> Configuration menu.

I think it makes sense to keep the same structure as https://github.com/OCA/account-invoice-reporting/pull/81

What do you think?

xavierjimenez:
Hi @HaraldPanten, 

It definitely makes sense, in fact, it was already done in the 10.0 version but the parent menu changed on 11.0, so that's why you couldn't see it. Now I've fixed it so it should be good.

Thanks for the review!

HaraldPanten (approved):
@xavierjimenez , now it's perfect :+1:

pedrobaeza:
I have fixed and completed this one as well, so merging

PR #46, opened 2018-03-20
[IMP] sale_comment_template: add partner template

cubells:
Add missing feature from older version: now you can assign a comment template to partner.
- Update spanish translation

cc @Tecnativa

WIP: Depends on https://github.com/OCA/account-invoice-reporting/pull/71

PR #56, opened 2018-10-18
[IMP] sale_comment_template: add partner template

cubells:
Related: https://github.com/OCA/sale-reporting/pull/46

Add missing feature from older version: now you can assign a comment template to partner.

* Update spanish translation

cc @Tecnativa


pedrobaeza (approved):
As this is direct forward-porting of a PR in v10, and everything is green, I merge.

PR #57, opened 2018-10-18
[11.0][MIG] sale_proforma_report

LoisRForgeFlow:
Migration to v11.


pedrobaeza:
v11 already has this as a native feature, so no need of migrating this. You only need to check the corresponding setting.

LoisRForgeFlow:
Ups, didn't know. Thanks for pointing it!!

pedrobaeza:
It's good to look at the migration issue where it was noted that already. What you can do is to include the OpenUpgrade merge mechanism, and look for a way to auto-add the group to all if you had previously installed that module (maybe checking the existence of an XML-ID).

LoisRForgeFlow:
@pedrobaeza started the work here: https://github.com/OCA/OpenUpgrade/pull/1572

PR #59, opened 2018-10-29
[MIG] sale_reporting_weight: Migration to 11.0

chienandalu:
This module adds another possible source to the sales report weight value:
 * If the UoM of the product is one of the category "Weight", the value is taken
  from the ordered quantity.

cc @Tecnativa


pedrobaeza:
This one can be closed due to new approach, isn't it?

pedrobaeza:
Closing in favor of https://github.com/OCA/product-attribute/pull/404

PR #61, opened 2018-12-11
sale_layout_category_hide_detail: New module

ernestotejeda:
Hide lines of specific sections in Quotations / Sale Order
and Invoice PDF report and in the website

Cc @Tecnativa


sale_layout_category_hide_detail/README.rst
```diff
@@ -0,0 +1,105 @@
...
+
+#. Go to *Sales -> Configuration -> Sales Orders -> Report Layout Categories*.
+#. Create a new *Report Layout Category* with *Hide detail* field on *True*
+   and create another one with *Hide detail* field on *False*.
+#. Go to *Sales -> Orders -> Quotations* and create a new *Quotation*.
+#. Add to the *Quotation* some lines associated with the first
+   *Report Layout Category* created before and add other lines associated with
+   the second *Report Layout Category* created.
+#. Print this Quotation. In the PDF report the section with *Hide detail*
+   field on *True* with be shown without its lines and with the subtotal.
```
pedrobaeza: s/with/will

sale_layout_category_hide_detail/README.rst
```diff
@@ -0,0 +1,105 @@
...
+#. Go to the Quotation in the website. The section with *Hide detail* field
+   on *True* with be shown without its lines and with the subtotal.
```
pedrobaeza: s/with/will

sale_layout_category_hide_detail/README.rst
```diff
@@ -0,0 +1,105 @@
...
+
+The behavior described before is the same for Quotations and  Invoices.
```
pedrobaeza: Extra space between `and` and `Invoices`.

sale_layout_category_hide_detail/README.rst
```diff
@@ -0,0 +1,105 @@
...
+#. Go to *Sales -> Configuration -> Sales Orders -> Report Layout Categories*.
+#. Create a new *Report Layout Category* with *Hide detail* field on *True*
+   and create another one with *Hide detail* field on *False*.
+#. Go to *Sales -> Orders -> Quotations* and create a new *Quotation*.
+#. Add to the *Quotation* some lines associated with the first
+   *Report Layout Category* created before and add other lines associated with
+   the second *Report Layout Category* created.
+#. Print this Quotation. In the PDF report the section with *Hide detail*
+   field on *True* with be shown without its lines and with the subtotal.
+#. Go to the Quotation in the website. The section with *Hide detail* field
```
pedrobaeza: s/website/customer portal

sale_layout_category_hide_detail/__manifest__.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2018 Tecnativa - Ernesto Tejeda
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Sale layout category hide detail",
+    "summary": "Hide lines of specific sections in Quotations /Sale Order and "
```
pedrobaeza: Switch to `Hide details for sections in sales orders and invoices for reports and customer portal`

sale_layout_category_hide_detail/__manifest__.py
```diff
@@ -0,0 +1,26 @@
...
+               "Invoice PDF report and in the website",
+    "version": "11.0.1.0.0",
+    "category": "Sales Management",
+    "website": "http://github.com/OCA/sale-workflow",
```
pedrobaeza: sale-reporting

sale_layout_category_hide_detail/models/sale_order.py
```diff
@@ -0,0 +1,30 @@
...
+
+from odoo import api, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    @api.multi
+    def lines_layouted_hide_detail(self, report_pages):
+        if self.user_has_groups('sale.group_sale_layout'):
```
pedrobaeza, line 12: Why do you need to check this here? Isn't this method directly not called when you don't have this group?

sale_layout_category_hide_detail/README.rst
```diff
@@ -0,0 +1,105 @@
...
+
+#. Go to *Sales -> Configuration -> Settings*
+#. Check *Sections on Sales Orders* option.
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to *Sales -> Configuration -> Sales Orders -> Report Layout Categories*.
```
pedrobaeza, line 53: This is also part of the configuration.

sale_layout_category_hide_detail/README.rst
pedrobaeza: Make clearer that there's no sense to not put subtotal when hiding details, so it doesn't matter if you check the "Add subtotal" or not. I would say more: you should hide that check if you mark "Hide detail", so:

1. Add this note here in the README.
2. Put "Hide detail" before "Add subtotal".
3. Hide "Add subtotal" when "Hide detail" is marked.
4. Rephrase to "Hide details".

sale_layout_category_hide_detail/views/sale_order_report_templates.xml
```diff
@@ -0,0 +1,20 @@
...
+			<td t-if="layout_category['hide_detail']" colspan="4" class="text-right" style="border-bottom: 1px solid black;">
+				<strong>Subtotal: </strong>
+				<span t-esc="layout_category['hide_detail_line_subtotal']" t-options="{'widget': 'monetary', 'display_currency': doc.pricelist_id.currency_id}"/>
+			</td>
+		</xpath>
+		<xpath expr="//tbody//span[@t-esc='subtotal']/../.." position="attributes">
+			<attribute name="t-if">not layout_category['hide_detail']</attribute>
+		</xpath>
+  	</template>
+</odoo>
```
pedrobaeza: Configure your IDE to always put a line at the end to avoid this.

sale_layout_category_hide_detail/views/sale_portal_templates.xml
```diff
@@ -0,0 +1,61 @@
...
+                                <span t-field="ol.price_subtotal" t-options='{"widget": "monetary", "display_currency": order.pricelist_id.currency_id}'/>
+                            </div>
+                        </div>
+                    </t>
+
+                    <t t-if="(layout_category_size > 1 or page_size > 1) and layout_category['subtotal'] and not layout_category['hide_detail']" groups="sale.group_sale_layout">
+                        <div class="row orders_vertical_align text-right">
+                            <div class="col-sm-12 col-xs-12">
+                                <strong>Subtotal: </strong>
+                                <t t-set="subtotal" t-value="sum(line.price_subtotal for line in layout_category['lines'])"/>
```
pedrobaeza: Isn't the value already computed in the method?

sale_layout_category_hide_detail/views/sale_portal_templates.xml
```diff
@@ -0,0 +1,61 @@
...
+                                <span t-esc="subtotal" t-options="{'widget': 'monetary', 'display_currency': order.pricelist_id.currency_id}"/>
```
pedrobaeza: Although this seems correctly formatted without bold, it appears bold any way:

seleccion_015

It happens in PDF report and in the portal. Can you check why isn't regular instead bold as the other subtotal?

sale_layout_category_hide_detail/views/account_portal_templates.xml
```diff
@@ -0,0 +1,55 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2018 Tecnativa - Ernesto Tejeda
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <template id="portal_invoice_report_layouted_hide_detail" inherit_id="account.portal_invoice_report">
+        <xpath expr="//span[@t-field='l.name']/../../.." position="replace">
```
pedrobaeza: Please NEVER do a replace. You should hide elements you don't want, or modify other attributes, but this is a total enemy of compatibility between several modules. Imagine there's other module that adds an extra column to each line. With this, you are removing the possibility of showing it.

sale_layout_category_hide_detail/views/invoice_report_templates.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2018 Tecnativa - Ernesto Tejeda
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <template id="report_invoice_layouted_hide_detaild" inherit_id="sale.report_invoice_layouted">
+        <xpath expr="//t[@name='lines_layouted']//tr[1]/td" position="replace">
```
pedrobaeza: NEVER replace.

sale_layout_category_hide_detail/views/sale_order_report_templates.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0"?>
+<!-- Copyright 2018 Tecnativa - Ernesto Tejeda
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <template id="report_saleorder_document_hide_detail" inherit_id="sale.report_saleorder_document">
+		<xpath expr="//tbody//tr[1]/td" position="replace">
```
pedrobaeza: NEVER replace.

sale_layout_category_hide_detail/views/sale_portal_templates.xml
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2018 Tecnativa - Ernesto Tejeda
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <template id="portal_order_page_layouted_hide_detail" inherit_id="sale.portal_order_page">
+        <xpath expr="//span[@t-field='ol.product_image']/../../.." position="replace">
```
pedrobaeza: NEVER replace.

pedrobaeza (changes requested):

pedrobaeza:
Other thing: remove the arrow for indicating if unfolded or not. We don't want to give that hint to the customers.

sale_layout_category_hide_detail/models/sale_order.py
ernestotejeda: This method is always called it doesn't matter if you have the group or not. From the qweb report this method is called and it always returns the lines grouped by categories, but the category's name is printed in the report if you have the group. The same happens with the subtotal.

pedrobaeza: OK

sale_layout_category_hide_detail/views/account_portal_templates.xml
```diff
@@ -0,0 +1,59 @@
...
+                                <td t-att-colspan="3 if layout_category['hide_details'] else 7" style="font-weight: bold; border-bottom: 1px solid black;">&amp;bull;
+                                    <t t-esc="layout_category['name']"/>
+                                </td>
+                                <td t-if="layout_category['hide_details']" colspan="4" class="text-right" style="border-bottom: 1px solid black;">
+                                    <strong>Subtotal: </strong>
+                                    <span t-esc="lines_subtotal" t-options="{'widget': 'monetary', 'display_currency': invoice.currency_id}"/>
+                                </td>
+                            </tr>
+                        </t>
+                        <tr t-foreach="layout_category['lines']" t-as="l">
```
pedrobaeza: This is still replacing previous layout. It doesn't matter you hide the other element instead of doing `position="replace"`. The result is the same. I ask about the same problem: what happens if other module adds an extra column to each order/invoice line and show it in the portal? with this, you will lose that column. Please do the same as with reports, changing only the minimum pieces of layout.

This is applicable to whole layout, although I put the comment here in the middle.

sale_layout_category_hide_detail/views/sale_portal_templates.xml
```diff
@@ -0,0 +1,63 @@
...
+                                <t t-esc="layout_category['name']"/>
+                            </div>
+                            <div t-if="layout_category['hide_details']" class="col-sm-8 col-xs-8 text-right">
+                                <strong>Subtotal: </strong>
+                                <span t-esc="lines_subtotal" t-options="{'widget': 'monetary', 'display_currency': order.pricelist_id.currency_id}"/>
+                            </div>
+                        </div>
+                    </t>
+
+                    <t t-foreach="layout_category['lines']" t-as="ol">
```
pedrobaeza: The same with this.

pedrobaeza (changes requested):

rafaelbn:
Tested in runbot just hide section if section is not used

- When "hide" is activated in the section it hide, yes, but in the PDF it must show the line as a normal line (white font,....)
- If no section is defined the the section mustn't be shown in the PDF (don't show uncategorized)

rafaelbn (dismissed):
Tested functionally in runbot :+1:

rafaelbn:
Sorry, Testing deeply in runbot with the Portal when using "hide" option the products are completly hiden and the section is also missing (we should show section)

Make a Sales order to a contact and after that give access to the contacto to portal, login as a customer in the portal

ernestotejeda:
> Tested in runbot just hide section if section is not used
> 
> * When "hide" is activated in the section it hide, yes, but in the PDF it must show the line as a normal line (white font,....)
> * If no section is defined the the section mustn't be shown in the PDF (don't show uncategorized)
> 

@rafaelbn, I don't understand The first point very well. What do you mean by 'show the line as a normal line (white font, ....)', I don't know if it refers to set a white background to the row where the name of the section is.
Is the intention that the section looks like a product? If this is the idea, I guess we should remove the word Subtotal in that row, right? In any case, we would have no image, quantity, unit price, and Taxes.

The word 'Uncategorized' comes from 'sale' Odoo module for these cases, anyway I can replace that row by a blank line.

sale_layout_category_hide_detail/views/account_portal_templates.xml
ernestotejeda: done, thank you

sale_layout_category_hide_detail/views/sale_portal_templates.xml
ernestotejeda: done, thank you

ernestotejeda:
@rafaelbn, https://github.com/OCA/sale-reporting/pull/61#issuecomment-449443261 is resolved

pedrobaeza (approved):

rafaelbn (approved):

PR #65, opened 2019-01-21
WIP [10.0][ADD] New module company_sale_note

acsonefho:
This module add a note field on the company who'll be printed on every sale reports.
This can be printed on different places:

- Before the sale note;
- At the end of the document (after fiscal position note).


pedrobaeza:
This seems quite similar to `base_comment_template` + derivative modules. You just need to tweak default value for this template and it will be done.

PR #66, opened 2019-04-01
[10.0][FIX] avoid reports with empty rows

pedrocs-exo:
This is a fix to avoid having empty rows on the QWeb sale order
report that relate to the formatted_note field. If the field has no
content, it shouldn't be displayed. Also note that the default
formatted_note content is <p><br></p> and so we also need to prevent
printing when the field has that value.

### DIFF
--- a/sale_comment_template/views/report_saleorder.xml
+++ b/sale_comment_template/views/report_saleorder.xml
@@ -17,17 +17,17 @@
         </p>
       </xpath>
       <xpath expr="//t[@t-foreach='page']/t[@t-as='l']" position="inside">
-        <t t-if="l.formatted_note">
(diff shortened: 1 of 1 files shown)


pedrobaeza:
But that are not empty rows... They have some HTML content. How did you come to this situation?

pedrocs-exo:
The default for html fields is:

`<p><br></p>`

I have that value on every order line. It certainly is html content but nevertheless is not something that the user entered so it should be considered as a form of "empty".

pedrobaeza:
OK, I see that indeed that default value is added, but please don't reindent the XML. Just change the affected line.

rousseldenis:
@pedrocasi

PR #70, opened 2019-07-05
[12.0][MIG] sale_layout_category_hide_detail

ernestotejeda:
Main changes:

  * Add JS (Add widget `boolean_fa_icon`, Add `show_in_line_section` option on fields statements in list views)
  * Adapt reports inheritance
  * Inherit `Sale order` form vew and `Invoice` form view to add fields in section lines

Cc @Tecnativa


sale_layout_category_hide_detail/README.rst
```diff
@@ -14,56 +14,56 @@ Sale layout category hide detail
...
+    :target: https://runbot.odoo-community.org/runbot/148/12.0
     :alt: Try me on Runbot
 
 |badge1| |badge2| |badge3| |badge4| |badge5| 
 
-This module allows you to set a new 'Hide details' boolean field on True in
-*Report Layout Categories* if you want to hide its lines of
-*Quotations / Sales Orders* and *Invoices* in the PDF report and in the
-website.
+This module allows you to set the new 'Show details' and 'Show subtotal'
```
pedrobaeza: Change redaction to:
```
This module allows you to check/unchek the new 'Show details' and 'Show subtotal' flags in *Sale order... if you want to show or hide their lines or their subtotals in...
```

sale_layout_category_hide_detail/README.rst
```diff
@@ -14,56 +14,56 @@ Sale layout category hide detail
...
-#. Add to the *Quotation* some lines associated with the first
-   *Report Layout Category* created before and add other lines associated with
-   the second *Report Layout Category* created.
-#. Print this Quotation. In the PDF report the section with *Hide details*
-   field on *True* will be shown without its lines and with the subtotal.
-#. Go to the Quotation in the customer portal. The section with *Hide details*
-   field on *True* will be shown without its lines and with the subtotal.
+#. In *Order lines* tab add a section line.
+#. Add some product lines.
+#. Add another section line, but this time, click on the *plus-circle* icon on
```
pedrobaeza: Maybe you can put here inline (`figure`?) the icon.

sale_layout_category_hide_detail/__manifest__.py
```diff
@@ -15,11 +15,11 @@
         "sale_management",
     ],
     "data": [
-        "views/sale_layout_category_view.xml",
-        "views/sale_order_report_templates.xml",
-        "views/sale_portal_templates.xml",
```
pedrobaeza: Why have you remove portal templates? Aren't they needed?

pedrobaeza (changes requested):
The option subtotal shown when details are shown:

Selección_035

would be better having the name of the section on the subtotal:

Selección_038

There aren't now any tests also...

pedrobaeza:
Portal must be modified as well:

Selección_039

sale_layout_category_hide_detail/views/sale_views.xml
```diff
@@ -0,0 +1,71 @@
...
+                                'icon_false': 'fa-eye-slash',
+                            },
+                            'terminology': {
+                                'hover_true': 'Switch to: details hidden',
+                                'hover_false': 'Switch to: details shown',
+                            }
+                       }"
+                       attrs="{'invisible': [('display_type', '!=', 'line_section')]}"/>
+                <field name="show_subtotal"
+                       widget="boolean_fa_icon"
```
yajo, line 34: I have some questions about this new widget.

1. Have you considered using `boolean_toggle` widget? I think it should be enough for your use case: 
Peek 09-07-2019 12-58

   The fact that it's not an eye doesn't seem to impact much in user experience IMHO.

1. In case it's not enough, better call it `boolean_icon`. In the future we might use a different icon set, so it's better to keep things unbranded if possible.

sale_layout_category_hide_detail/views/sale_views.xml
```diff
@@ -0,0 +1,71 @@
...
+                       options="{
+                            'show_in_line_section': true,
+                            'fa_icons': {
+                                'icon_true': 'fa-plus-circle',
+                                'icon_false': 'fa-minus-circle',
+                            },
+                            'terminology': {
+                                'hover_true': 'Switch to: subtotal hidden',
+                                'hover_false': 'Switch to: subtotal shown',
```
yajo, line 43: You didn't consider translations here. Once you fix it as explained in https://github.com/OCA/sale-reporting/pull/70#discussion_r302400298, you'll have to change this to maybe `'terminology': 'show'`

sale_layout_category_hide_detail/static/src/js/boolean_fa_icon_widget.js
```diff
@@ -0,0 +1,66 @@
+/* Copyright 2019 Tecnativa - Ernesto Tejeda
+ * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+ */
+odoo.define('sale_layout_category_hide_detail.boolean_fa_icon_widget', function (require) {
+    "use strict";
+
+    var AbstractField = require('web.AbstractField');
+    var registry = require('web.field_registry');
+
+    var BooleanFaIconWidget = AbstractField.extend({
```
yajo, line 13: Please check the other comments on a consideration if this is really needed.

If we see it's really needed, then:

1. Here it would make more sense to inherit from `FieldBoolean`, [just as `boolean_toggle` widget does to simplify its code](https://github.com/odoo/odoo/blob/674ee035ca954a57341cbaa951d2f6f31b5f8fe2/addons/web/static/src/js/fields/basic_fields.js#L2143-L2164).
1. The widget should go in a separate module in OCA/web IMHO.

sale_layout_category_hide_detail/static/src/js/boolean_fa_icon_widget.js
```diff
@@ -0,0 +1,66 @@
...
+        /**
+         * Render font-awesome icon based on state
+         *
+         * @override
+         * @private
+         */
+        _render: function () {
+            var options = this.attrs.options;
+            var fa_class = this.value ? options.fa_icons.icon_true : options.fa_icons.icon_false;
+            var tip = this.value ? options.terminology.hover_true : options.terminology.hover_false;
```
yajo: Again, if we really are going to keep this widget, these things need default values.

yajo (changes requested):

sale_layout_category_hide_detail/views/sale_views.xml
pedrobaeza: About the `boolean_toggle`, I prefer this icon set without column name, as it would require a lot of width space. That's the main reason for this. Maybe we can extract the widget in OCA/web...

yajo: OK, I just did a quick test and this is how it looks:

Captura de pantalla de 2019-07-09 13-41-01

I can see the difference, indeed.

Then we need to handle:

1. Translations. Important!
1. Extract into OCA/web.

pedrobaeza: Let's put this in the ROADMAP as we don't need this widget for other things and this task has required too much time, so we shouldn't increase it with this extraction.

Yajo: OK, but at least the translations issue should be fixed IMHO. Easier to do if not extracting the widget, BTW.

yajo: OK, seems a good solution.

ernestotejeda:
> The option subtotal shown when details are shown:

@pedrobaeza I kept the same behavior as in the previous version: The option subtotal shown when details are shown and when you uncheck show details, show subtotal option is hide

sale_layout_category_hide_detail/__manifest__.py
ernestotejeda: The portal sale order template was removed wrongly, but regarding to the portal invoice, now in v12 there isn't portal invoice template, odoo use the same template of the PDF report. That template is rendered and shown in a <frame>. So, fortunately we don't have to redefine a portal invoice template :)

sale_layout_category_hide_detail/views/sale_views.xml
ernestotejeda: boolean_toggle is a FieldBoolean with the checkbox made up with scss, but this new widget is not rendered as a checkbox and its behavior is not the same as the FieldBoolean. If I inherited from a FieldBoolean I would have to rewrite almost all FieldBoolean methods to disable its behavior, and then the inheritance would no longer make sense. It is the same as what happens with FieldBooleanButton, so that inherits from AbstractField instead of FieldBoolean

sale_layout_category_hide_detail/static/src/js/boolean_fa_icon_widget.js
ernestotejeda: - boolean_toggle is a FieldBoolean with the checkbox made up with scss, but this new widget is not rendered as a checkbox and its behavior is not the same as the FieldBoolean. If I inherited from a FieldBoolean I would have to rewrite almost all FieldBoolean methods to disable its behavior, and then the inheritance would no longer make sense. It is the same as what happens with FieldBooleanButton, so that inherits from AbstractField instead of FieldBoolean
- regarding to have the widget in a separated module I agree with Pedro (https://github.com/OCA/sale-reporting/pull/70#discussion_r301590394)

ernestotejeda:
Changes Done

sale_layout_category_hide_detail/static/src/js/boolean_fa_icon_widget.js
```diff
@@ -0,0 +1,78 @@
...
+         */
+        _render: function () {
+            //set icon class
+            var fa_icons = this.attrs.options.fa_icons;
+            var icon_true = fa_icons && fa_icons.icon_true || 'fa-check-square-o';
+            var icon_false = fa_icons && fa_icons.icon_false || 'fa-square-o';
+            var fa_class = this.value ? icon_true : icon_false;
+            //set tip message
+            var terminology = this.attrs.options.terminology;
+            var hover_true = terminology && _(terminology.hover_true) || _('Click to uncheck');
```
yajo: In JS, `_()` is the [underscore.js](http://underscorejs.org/) library, not the translation function. To translate, it's `_t()`.

OTOH, [the only legal content of a `_t()` is always a raw string](https://www.odoo.com/documentation/12.0/reference/translations.html#explicit-exports). You cannot use variables within, because the contents are processed statically to produce the `.pot` file.

But you cannot translate from the `options` attribute when declaring the widget in the XML view file. How to work it around?

I imagine you got confused because I told you to look at some code that turned out to be buggy, as you can see in https://github.com/odoo/odoo/pull/34774.

What can you do here? `this.nodeOptions.terminology` should be a key from a frozen set of choices, like "check" and "show". [Then, the code, similar to what the `FieldBooleanButton` widget does here](https://github.com/odoo/odoo/blob/04d9715766adab97dd6da66c97e50fd6a30dab31/addons/web/static/src/js/fields/basic_fields.js#L2112-L2124).

ernestotejeda: @Yajo , check what i did and tell me what you think. I have done a separate 'Commit' with some changes to be able to translate the tooltip messages in order to make it more customizable. I know it works but I want your opinion.

sale_layout_category_hide_detail/tools/translate.py
```diff
@@ -0,0 +1,6 @@
+# Copyright 2018-2019 Tecnativa - Ernesto Tejeda
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo.tools.translate import TRANSLATED_ATTRS
+
+TRANSLATED_ATTRS |= {'hover_true', 'hover_false'}
```
yajo: This will get executed even if the addon is not installed.

sale_layout_category_hide_detail/static/src/js/boolean_fa_icon_widget.js
yajo: It's got a bad detail that I commented below in the line.

Apart from that, event when this works fine and is a good solution, I still think that [doing this](https://github.com/odoo/odoo/blob/04d9715766adab97dd6da66c97e50fd6a30dab31/addons/web/static/src/js/fields/basic_fields.js#L2112-L2124) will be better for our use case. It's just more standard code, and for our case at hand it's enough to get the job done, so I think it's better not add any more complexity than the one needed for now.

ernestotejeda:
Changes done, @Yajo check it out please

yajo (approved):
Much simpler. Thanks!

ernestotejeda:
@pedrobaeza check it out

pedrobaeza (changes requested):
Go for example to

http://3382265-70-eee420.runbot2-2.odoo-community.org/web#id=7&action=258&model=sale.order&view_type=form&menu_id=139

If you both print or go to portal, you don't see any lines. That's because they don't have any section involved. You must take that into account as well.

ernestotejeda:
Thanks @pedrobaeza , even in the invoice report and in the portal invoice the lines were also not shown when there wasn't at least one section line. But I have already fixed it.

pedrobaeza (approved):
Everything seems fine now. I have added the needed migration script when coming from previous version

pedrobaeza:
@ernestotejeda I have just found that you lost the file account_portal_templates.xml in the migration. I'm restoring it.

pedrobaeza:
OK, seen that in invoices, the HTML is directly rendered from the same report than the PDF.

PR #73, opened 2019-10-21
[12.0][FIX] sale_comment_template: unknown res.partner attribute

ernestotejeda:
Cc @Tecnativa


pedrobaeza (approved):
Fast-tracking this as obvious after refactoring in https://github.com/OCA/account-invoice-reporting/pull/107

PR #74, opened 2019-11-12
[12.0][FIX] sale_layout_category_hide_detail: Propagate 'Show details' to invoice"

ernestotejeda:
Propagate values in 'Show details' and 'Show subtotal' fields from Sale Order Lines to Invoice Lines

Cc @Tecnativa TT18052

### DIFF
--- a/sale_layout_category_hide_detail/models/sale_order.py
+++ b/sale_layout_category_hide_detail/models/sale_order.py
@@ -13,3 +13,9 @@ class SaleOrderLine(models.Model):
     show_subtotal = fields.Boolean(
         string="Show subtotal",
         default=True)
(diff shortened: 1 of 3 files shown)


sale_layout_category_hide_detail/models/sale_order.py
```diff
@@ -13,3 +13,10 @@ class SaleOrderLine(models.Model):
     show_subtotal = fields.Boolean(
         string="Show subtotal",
         default=True)
+
+    @api.multi
```
pedrobaeza: No need of `@api.multi` since v10

pedrobaeza (approved):

PR #78, opened 2019-11-20
[12.0][WIP]Attempt to insert page-break option

Vartkat:
As I'm not really a devel, that's as far I could go. 
I think I'm I used the right way to do things but I need comment and help from pros.

Thanks for helping

Regards


pedrobaeza:
@guewen @TDu maybe you are interested in this.

@Vartkat there are now conflicts.

PR #82, opened 2019-12-02
[MIG][12.0] sale_report_delivered_subtotal: Migration to v12.0

sergio-teruel:
cc @Tecnativa TT20186
@chienandalu @Yajo Can you review?


sale_report_delivered_subtotal/reports/sale_report.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class SaleReport(models.Model):
+    _inherit = "sale.report"
+
+    price_subtotal_delivered = fields.Float(
+        string='Total Price Delivered',
+    )
+
+    def _query(self, with_clause='', fields={}, groupby='', from_clause=''):
```
pedrobaeza: For avoiding the linter (which is correct that can be dangerous), but with this change it doesn't affect anything:
```suggestion
    def _query(self, with_clause='', fields=None, groupby='', from_clause=''):
```

sale_report_delivered_subtotal/reports/sale_report.py
```diff
@@ -0,0 +1,26 @@
...
+        fields = dict(fields or {})
```
pedrobaeza: Adapted to the above code:
```suggestion
        if fields is None:
            fields = {}
```

pedrobaeza (changes requested):
You can also squash administrative commits with the first non-administrative one:

Selección_027

sale_report_delivered_subtotal/reports/sale_report.py
yajo: Actually it would work as it is.

pedrobaeza: It's for not creating a new one if not needed and then build a new dict for it, but it's not critical indeed.

sergio-teruel:
Changes done!!

pedrobaeza (approved):

carlosdauden (approved):

PR #84, opened 2020-01-28
[ADD][11.0] sale_report_supplier_filter: New module

chienandalu:
This module allows to filter by product supplier in the sale reports to obtain
the sales for that specific set of products.

cc @Tecnativa TT21143

### DIFF
--- a/sale_report_supplier_filter/reports/report_all_channels_sales.py
+++ b/sale_report_supplier_filter/reports/report_all_channels_sales.py
@@ -1,0 +1,12 @@
+# Copyright 2020 Tecnativa - David Vidal
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class ReportAllChannelsSales(models.Model):
+    _inherit = "report.all.channels.sales"
+
+    seller_id = fields.Many2one(
+        related='product_id.seller_ids.name',
+        domain=[('supplier', '=', True)],
+    )
--- a/sale_report_supplier_filter/__init__.py
+++ b/sale_report_supplier_filter/__init__.py
@@ -1,0 +1,1 @@
+from . import reports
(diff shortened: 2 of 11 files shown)


rafaelbn:
@chienandalu please could you check travis? runbot is not available. Thanks! looks great! :smile:

pedrobaeza:
Please don't use GH conflict resolution or add merge commits as this mess up commit history.

chienandalu:
Sorry, I pushed before rebase

sale_report_supplier_filter/reports/report_all_channels_sales.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2020 Tecnativa - David Vidal
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+from odoo import fields, models
+
+
+class ReportAllChannelsSales(models.Model):
+    _inherit = "report.all.channels.sales"
+
+    seller_id = fields.Many2one(
+        related='product_id.seller_ids.name',
```
rousseldenis, line 10: IMHO, you have to say in README that this is applicable for main supplier at time of report, if the main supplier was another one at sale time, the result could be incorrect. Or I misunderstand the aim of this?

rafaelbn (approved):
:+1:  tested in production

sale_report_supplier_filter/reports/report_all_channels_sales.py
rafaelbn: Hello @rousseldenis , it shows of **any** supplier , nor the main. It is intended and discussed with different customers. All customer ask for the same thing, **any**.

carlosdauden (approved):
It is very useful because it allows you to filter by any provider that is established in the product supplierinfo, but as @rousseldenis says, you can comment in the README that if the field is shown, only the first in the list (at  time of report) will appear.

PR #87, opened 2020-03-20
[12.0][FIX] sale_layout_category_hide_detail: Fix error gotten in tree with buttons

ernestotejeda:
Before this PR:
If there are buttons in the tree we get an error

Cc @Tecnativa TT22935

### DIFF
--- a/sale_layout_category_hide_detail/readme/USAGE.rst
+++ b/sale_layout_category_hide_detail/readme/USAGE.rst
@@ -28,18 +28,18 @@ following:
 
 The behavior described before is the same for Quotations and Invoices.
 
-.. |eye-icon| image:: sale_layout_category_hide_detail/static/description/readme-icons/eye.png
+.. |eye-icon| image:: ../static/description/readme-icons/eye.png
    :alt: plus-circle icon
    :width: 12 px
 
-.. |eye-slash-icon| image:: sale_layout_category_hide_detail/static/description/readme-icons/eye-slash.png
+.. |eye-slash-icon| image:: ../static/description/readme-icons/eye-slash.png
    :alt: minus-circle icon
    :width: 12 px
 
-.. |plus-circle-icon| image:: sale_layout_category_hide_detail/static/description/readme-icons/plus-circle.png
+.. |plus-circle-icon| image:: ../static/description/readme-icons/plus-circle.png
    :alt: plus-circle icon
    :width: 12 px
 
-.. |minus-circle-icon| image:: sale_layout_category_hide_detail/static/description/readme-icons/minus-circle.png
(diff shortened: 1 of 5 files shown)


sale_layout_category_hide_detail/readme/USAGE.rst
```diff
@@ -28,18 +28,18 @@ following:
 
 The behavior described before is the same for Quotations and Invoices.
 
-.. |eye-icon| image:: sale_layout_category_hide_detail/static/description/readme-icons/eye.png
+.. |eye-icon| image:: static/description/readme-icons/eye.png
```
pedrobaeza: Please change to

```suggestion
.. |eye-icon| image:: ../static/description/readme-icons/eye.png
```

for getting a good visualization on each possible file.

sale_layout_category_hide_detail/static/src/js/sale_layout_category_hide_detail.js
```diff
@@ -37,16 +39,25 @@ odoo.define('sale_layout_category_hide_detail.sale_layout_category_hide_detail',
...
                     section_fields_count ++;
             });
             return section_fields_count;
         },
         _renderHeaderCell: function (node) {
             var $th = this._super.apply(this, arguments);
-            var options = this.state.fieldsInfo.list[node.attrs.name].options;
-            var show_in_line_section = options.show_in_line_section;
+
+            var show_in_line_section = false;
```
pedrobaeza: @Tardo can you help in a more compact form of putting this?

pedrobaeza (changes requested):

sale_layout_category_hide_detail/static/src/js/sale_layout_category_hide_detail.js
Tardo: Something like this?
```
var field_info = this.state.fieldsInfo.list[node.attrs.name];
var show_in_line_section = field_info && field_info.options.show_in_line_section;
$th.text("").toggleClass('o_column_sortable', !show_in_line_section);
```

pedrobaeza: Uhm, not sure. `o_column_sortable` appears in the rest of the cases, @ernestotejeda ?

ernestotejeda: Yes, I just want to remove o_column_sortable class  in case show_in_line_section is true

ernestotejeda:
I reduced the code a bit according to what @Tardo suggested

ernestotejeda:
> I reduced the code a bit according to what @Tardo  suggested

pedrobaeza (approved):

PR #89, opened 2020-04-02
[12.0][IMP][sale_layout_category_hide_detail] New feature to hide/show detail prices

ivantodorovich:
Implements a "Show prices" feature.

This is particularly usefull in travel industry, where multiple products are sold but only the final price is relevant.

In order to add this button without making the UX ugly, we've changed a bit the icons to use checkboxes instead, which are more intuitive.

Even though both features share the same "checkbox" button, it's easy to identify them beacuse of the tooltips.

icons

show_details

show_prices

show_subtotals

sample_config

sample_report


ivantodorovich:
ping @Tecnativa

sale_layout_category_hide_detail/views/account_invoice_view.xml
```diff
@@ -35,28 +37,43 @@
...
+                           }"
+                           attrs="{'invisible': ['|', ('display_type', '!=', 'line_section'), ('show_details', '!=', True)]}"/>
                     <field name="show_subtotal"
                            widget="boolean_fa_icon"
                            options="{
                                 'show_in_line_section': true,
                                 'fa_icons': {
-                                    'icon_true': 'fa-plus-circle',
-                                    'icon_false': 'fa-minus-circle',
+                                    'icon_true': 'fa-check-circle',
```
gregory-moka, line 64: Suggestion : fa-check-square

sale_layout_category_hide_detail/views/account_invoice_view.xml
```diff
@@ -35,28 +37,43 @@
...
+                                    'icon_false': 'fa-circle',
```
gregory-moka, line 65: Suggestion : fa-square

sale_layout_category_hide_detail/views/sale_views.xml
```diff
@@ -55,12 +72,15 @@
...
+                    <div class="col-3">
                         <span class="float-right text-right">
                             <span t-attf-class="ml-3 fa #{record.show_details.raw_value ? 'fa-eye' : 'fa-eye-slash'}"
                                   t-att-aria-label="record.show_details.raw_value ? 'Details shown' : 'Details hidden'"/>
                             <span t-if="record.show_details.raw_value"
-                                  t-attf-class="ml-3 fa #{record.show_subtotal.raw_value ? 'fa-plus-circle' : 'fa-minus-circle'}"
+                                  t-attf-class="ml-3 fa #{record.show_price.raw_value ? 'fa-check-circle' : 'fa-circle'}"
+                                  t-att-aria-label="record.show_price.raw_value ? 'Price shown' : 'Price hidden'"/>
+                            <span t-if="record.show_details.raw_value"
+                                  t-attf-class="ml-3 fa #{record.show_subtotal.raw_value ? 'fa-check-circle' : 'fa-circle'}"
```
gregory-moka, line 83: Suggestion : 
t-attf-class="ml-3 fa #{record.show_subtotal.raw_value ? 'fa-check-square' : 'fa-square'}"

ivantodorovich:
Forward-ported to 13.0 here: https://github.com/OCA/sale-reporting/pull/90/files

sale_layout_category_hide_detail/views/account_invoice_view.xml
ivantodorovich: Having different checkboxes icons didn't provide anything to the UX, that's why we're using the same icon for both features.

They can be easily identified by the Tooltip, and by their order :)

pedrobaeza: I don't like having the same icon.

ivantodorovich: Would `check-circle` / `check-square` be ok?

There aren't many choises to pick 

ie: 
show prices (square)
show subtotal (circle)

pedrobaeza: @ernestotejeda what do you think?

ernestotejeda: I agree with @pedrobaeza, different icons are one more resource for the user to more easily and quickly identify the option they need to check/uncheck (or activate/deactivate). Once the user learns what functionality each icon corresponds to (even if it is not the most appropriate), they can locate it in a single glance without having to hover the mouse or learn the order in which they are located.

ernestotejeda: Another thing that worries me is that I don't know if it's appropriate to add this functionality in this module. Perhaps the users who are using this module don't need this new functionality and it may be uncomfortable for them to have one more column that they don't use, or maybe they simply don't want one more.
Ping @pedrobaeza

ivantodorovich: @ernestotejeda what about using:

- Show prices: `square` / `check-square`: https://fontawesome.com/v4.7.0/icon/check-square
- Show subtotals: `circle` / `check-circle`: https://fontawesome.com/v4.7.0/icon/check-circle

I don't think we'll find toggleable descriptive icons for these features :sweat_smile: 

The ones that were used before this PR for show subtotals were `plus-circle` and `minus-circle` but, honestly, using `check-circle` is more intuitive as users already know that if it's checked it means `true`

---

Regarding moving this to a new module.. I understand that users might not want this feature but having separate modules it probably overkill, don't you think? Specially because we depend on Show detail, and features are clearly related.

Maybe this could be managed by security groups, to enable / disable the features

ivantodorovich: @ernestotejeda an update on this one?

gregory-moka: @ernestotejeda and @pedrobaeza : What do you think about that ?

gregory-moka: Works really fine, and users appreciate the options to show or hide what the need.
Nice job !
I'm waiting the PR validation to finish the translation on weblate (I did it for the actual version this afternoon).

PR #99, opened 2020-11-10
[FIX][12.0] group menu sale backorder missing

sergiocorato:
Description: without this group permission, menu is visible to everyone.
How to replicate: create an user without sale permission.
Before this PR: user see anyway Sale app menu, but only with print menu of SO Backorder.
After this PR: user do not see Sale app menu.

### DIFF
--- a/sale_backorder/wizard/so_backorder_wizard_view.xml
+++ b/sale_backorder/wizard/so_backorder_wizard_view.xml
@@ -32,5 +32,6 @@
     <menuitem id="menu_so_backorder_report"
               action="action_so_backorder_wizard"
               parent="sale.sale_order_menu"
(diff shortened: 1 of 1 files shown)


smangukiya (approved):

rousseldenis:
@sergiocorato I think you have a problem with your config.

Tested with demo user on runbot :

image

sergiocorato:
@rousseldenis I added a description. This group permission is needed to restrict the permission to user that have at least this permission (removing sale permission, an user will continue to view this menu).

sergiocorato:
@OCA/crm-sales-marketing-maintainers some opinion?

sergiocorato:
@tafaRU :pray:

PR #100, opened 2020-11-18
[NEW] sale_order_report_hide_line

Rad0van:
This is very similar to `account_invoice_report_hide_line` and serves to hide lines in SO that have unit price set to 0 to be able to include them in margin calculation for example but you do not want customers to see them.


Rad0van:
@pedrobaeza , @chienandalu , @ernestotejeda I did this last year. Does it make sense? Would it be OK to include it? There are some thing to fix for sure (description, ...). Will do if need be.

chienandalu (commented):
Well, indeed we've come accross this need with a customer in the past, although we considered it a custom requirement at the time.

But it seems not only them need such feature :slightly_smiling_face:

Rad0van:
I still think this could be usable for other people. This requirement came up recently with another project (although for 15.0).
@chienandalu or @pedrobaeza could you please add 'no stale' label and reopen it? Thank you.

PR #97, opened 2020-09-18
[13.0] [MIG] sale_comment_template: Module migration in v13.

Chandresh-SerpentCS:

PR #105, opened 2021-02-16
[13.0][MIG] sale_comment_template: Migration 13.0

victoralmau:
Supersedes https://github.com/OCA/sale-reporting/pull/97 to fix tests.

Locked by:
- [x] https://github.com/OCA/account-invoice-reporting/pull/174 `account_invoice_comment_template`
- [x] https://github.com/OCA/reporting-engine/pull/504

@Tecnativa TT26776


joao-p-marques:
Copier template applied to fix pre-commit @victoralmau https://github.com/OCA/sale-reporting/pull/106

sale_comment_template/tests/test_sale_order_report.py
```diff
@@ -0,0 +1,80 @@
...
+            self.sale_order.get_comment_template("after_lines"),
+            invoice.get_comment_template_item("after_lines").text,
+        )
+        res = (
+            self.env["ir.actions.report"]
+            ._get_report_from_name("account.report_invoice")
+            .render_qweb_html(invoice.ids)
+        )
+        self.assertRegexpMatches(str(res[0]), self.move_before_comment.text)
+        self.assertRegexpMatches(str(res[0]), self.move_after_comment.text)
```
ernestotejeda, line 81: The same I commented here https://github.com/OCA/account-invoice-reporting/pull/174#discussion_r608901885

sale_comment_template/tests/test_sale_order_report.py
```diff
@@ -0,0 +1,80 @@
...
+        )
+
+    def test_comments_in_sale_order(self):
+        res = (
+            self.env["ir.actions.report"]
+            ._get_report_from_name("sale.report_saleorder")
+            .render_qweb_html(self.sale_order.ids)
+        )
+        self.assertRegexpMatches(str(res[0]), self.sale_before_comment.text)
+        self.assertRegexpMatches(str(res[0]), self.sale_after_comment.text)
```
ernestotejeda: The same I commented here https://github.com/OCA/account-invoice-reporting/pull/174#discussion_r608901885

ernestotejeda (commented):
Why not keep the same functionality of being able to write a custom comment on each document?

sale_comment_template/tests/test_sale_order_report.py
victoralmau: IMO it's necessary to test it. To prevent errors in other addons to depend it i have removed `post_install`.

ernestotejeda: Thanks, I got an error in the tests of another module when it was tested together with this module, but that has been solved now that you have removed the post_install, so I think that check could be kept :)

sale_comment_template/tests/test_sale_order_report.py
```diff
@@ -1,41 +1,47 @@
 # Copyright 2017 Simone Rubino - Agile Business Group
 # Copyright 2018 Tecnativa - Pedro M. Baeza
+# Copyright 2021 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
 import odoo.tests
 from odoo.tests.common import TransactionCase
 
 
 @odoo.tests.common.at_install(False)
```
joao-p-marques: I think you also need to remove this line, otherwise the tests will be completely skipped. I am not sure, though, have you tested this and made sure the tests actually ran locally?
```suggestion
```

pedrobaeza:
This should be adapted now to the new many2many reality: the fields in the invoice should be also m2m, and filter them by company.

sale_comment_template/__manifest__.py
```diff
@@ -6,8 +6,9 @@
 {
     "name": "Sale Comments",
     "summary": "Comments texts templates on Sale documents",
-    "version": "12.0.1.1.0",
+    "version": "13.0.1.1.0",
```
joao-p-marques: Version should be `13.0.1.0.0`

joao-p-marques (changes requested):

victoralmau:
It's ready to review now, right?

sale_comment_template/migrations/13.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,31 @@
...
+@openupgrade.migrate()
+def migrate(env, version):
+    openupgrade.logged_query(
+        env.cr,
+        """
+        INSERT INTO sale_order_base_comment_template_rel
+        (base_comment_template_id, sale_order_id)
+        SELECT bct.id AS base_comment_template_id, so.id AS sale_order_id
+        FROM sale_order so
+        JOIN base_comment_template bct ON bct.id = so.old_comment_template1_ids
```
pedrobaeza: `old_comment_template1_ids` shouldn't be `comment_template1_id`? (and the same for the other)

pedrobaeza: Anyways, you don't need the JOIN to this table, just select `so.comment_template1_id` above.

sale_comment_template/views/base_comment_template_view.xml
```diff
@@ -1,9 +1,23 @@
...
+    <record model="ir.actions.act_window" id="action_sale_order_comment_template">
+        <field name="name">Sale Orders Comments</field>
+        <field name="type">ir.actions.act_window</field>
+        <field name="res_model">base.comment.template</field>
+        <field name="view_mode">tree,form</field>
+        <field
+            name="context"
+            eval="{'default_model_ids': [(4, ref('sale.model_sale_order'))]}"
+        />
+        <field
```
pedrobaeza, line 13: No domain here?

sale_comment_template/views/report_saleorder.xml
```diff
@@ -1,40 +1,24 @@
...
-        <p t-if="doc.note2">
-          <span t-field="doc.note2"/>
-        </p>
-      </xpath>
+    <template
+        id="report_saleorder_document_comments"
+        inherit_id="sale.report_saleorder_document"
+    >
+        <xpath expr="//table[hasclass('table-sm')]" position="before">
+            <p
```
pedrobaeza: Is this extra `<p>` correct that we insert inside the HTML raw? Shouldn't it be a `<div>` or similar?

pedrobaeza (changes requested):

sale_comment_template/migrations/13.0.1.0.0/post-migration.py
pedrobaeza: I have fixed myself this point for not having an error on upgrade.

victoralmau:
Thanks @pedrobaeza , changes done.

pedrobaeza (approved):

victoralmau:
It's ready to merge? I'm just asking because these changes is pending in 14.0?

chienandalu (approved):

sale_comment_template/tests/test_sale_order_report.py
```diff
@@ -1,74 +1,76 @@
...
+        self.partner = self.env["res.partner"].create({"name": "Partner Test"})
+        self.partner.base_comment_template_ids = [
+            (4, self.sale_before_comment.id),
+            (4, self.sale_after_comment.id),
+            (4, self.move_before_comment.id),
+            (4, self.move_after_comment.id),
+        ]
+        self.sale_order = self.env.ref("sale.sale_order_7").copy(
+            {"partner_id": self.partner.id, "partner_invoice_id": self.partner.id}
+        )
```
joao-p-marques: Can you create a specific sale order instead of copying? Or, if you copy, can you assign the proper company id? I think this might cause integration issues if not solved...

victoralmau: Thanks, changes done.

joao-p-marques (approved):

PR #107, opened 2021-03-09
[12.0][FIX] change partner do not update comment translation

sergiocorato:
Use case: set a comment template in SO directly or as a default. Do not set a default comment template in partner. If partner is changed, the comment remains the same, even if there are translations.

Before this PR: the comment is not updated on partner's language.

After this PR: the comment is updated on partner's language.


rousseldenis:
@sergiocorato Maybe a little test for that change?

PR #114, opened 2021-06-11
[13.0][ADD] sale_order_line_position

TDu:
This module adds an auto computed position on sale order line.
This position number is added on the report.

The position can be used to keep track of each line during
the delivery, invoicing of the order with the customer.
This is why there is related modules on `account-invoice-reporting`
and `stock-logisics-reporting`.

The position set on a line is not changed when the order is not in
draft anymore.

### DIFF
--- a/sale_order_line_position/models/sale_order.py
+++ b/sale_order_line_position/models/sale_order.py
@@ -1,0 +1,28 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import api, fields, models
+
+
+class SaleOrder(models.Model):
+    _inherit = "sale.order"
+
+    locked_positions = fields.Boolean(compute="_compute_locked_positions")
+
+    @api.depends("state")
+    def _compute_locked_positions(self):
+        for record in self:
+            record.locked_positions = record.state != "draft"
+
+    def action_quotation_send(self):
+        self.recompute_positions()
+        return super().action_quotation_send()
+
+    def recompute_positions(self):
+        self.ensure_one()
+        if self.locked_positions:
+            return
+        lines = self.order_line.filtered(lambda l: not l.display_type)
+        lines.sorted(key=lambda x: x.sequence)
+        for position, line in enumerate(lines, start=1):
+            line.position = position
--- a/sale_order_line_position/models/sale_order_line.py
+++ b/sale_order_line_position/models/sale_order_line.py
@@ -1,0 +1,68 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import api, fields, models
+
+
+class SaleOrderLine(models.Model):
+    _inherit = "sale.order.line"
+
+    position = fields.Integer(readonly=True, index=True, default=False)
+    position_formatted = fields.Char(compute="_compute_position_formatted")
+
+    @api.depends("position")
+    def _compute_position_formatted(self):
+        for record in self:
+            record.position_formatted = record._format_position(record.position)
+
+    @api.onchange("sequence")
+    def _onchange_sequence(self):
+        if self.order_id.locked_positions:
+            return
+        lines = self.order_id.order_line.filtered(
+            lambda x: not x.display_type and x.sequence < self.sequence
+        )
+        self.position = len(lines) + 1
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        vals_list = self._add_next_position_on_new_line(vals_list)
+        return super().create(vals_list)
+
+    def unlink(self):
+        sales = self.mapped("order_id")
+        res = super().unlink()
+        for sale in sales:
+            sale.recompute_positions()
+        return res
+
+    def _add_next_position_on_new_line(self, vals_list):
+        sale_ids = [
+            line["order_id"]
+            for line in vals_list
+            if not line.get("display_type") and line.get("order_id")
+        ]
+        if sale_ids:
+            ids = tuple(set(sale_ids))
+            self.flush()
+            query = """
+            SELECT order_id, max(position) FROM sale_order_line
+            WHERE order_id in %s GROUP BY order_id;
+            """
+            self.env.cr.execute(query, (ids,))
+            default_pos = {key: 1 for key in ids}
+            existing_pos = {
+                order_id: pos + 1 for order_id, pos in self.env.cr.fetchall()
+            }
+            sale_pos = {**default_pos, **existing_pos}
+            for line in vals_list:
+                if not line.get("display_type"):
+                    line["position"] = sale_pos[line["order_id"]]
+                    sale_pos[line["order_id"]] += 1
+        return vals_list
+
+    @api.model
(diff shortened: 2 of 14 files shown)


sale_order_line_position/tests/test_sale_order_line_position.py
```diff
@@ -0,0 +1,48 @@
...
+
+class TestSaleOrderLinePosition(SingleTransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))
+        cls.partner = cls.env.ref("base.res_partner_12")
+        cls.product = cls.env.ref("product.product_product_9")
+        cls.order = cls.env["sale.order"].create({"partner_id": cls.partner.id})
+
```
mmequignon: a test about what happens once the SO isn't in draft state would be nice

mmequignon (approved):

sale_order_line_position/tests/test_sale_order_line_position.py
TDu: @mmequignon Thanks for your review, I added/improved the tests and refactor a little.

sale_order_line_position/models/sale_order_line.py
```diff
@@ -0,0 +1,55 @@
...
+        if self.order_id.locked_positions:
+            return
+        lines = self.order_id.order_line.filtered(
+            lambda x: not x.display_type and x.sequence < self.sequence
+        )
+        self.position = len(lines) + 1
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        sale_ids = [
```
simahawk: can't we do this only on SO write of `order_line`?
Also, does it make sense to do it for all order states?
I think this is adding too much overhead at lines' creation.

sale_order_line_position/models/sale_order_line.py
```diff
@@ -0,0 +1,55 @@
...
+            line["order_id"]
+            for line in vals_list
+            if not line.get("display_type") and line.get("order_id")
+        ]
+        sale_pos = {}
+        sales = self.env["sale.order"].browse(sale_ids)
```
simahawk: instead of loading orders can't you just do a select for max position?

simahawk (changes requested):

sale_order_line_position/models/sale_order_line.py
simahawk: Or, can't we do this only when printing? (eg: override report handler, and/or via ctx key etc)

sale_order_line_position/models/sale_order_line.py
TDu: I did that on my last fixup

sale_order_line_position/models/sale_order_line.py
TDu: Can not be done at printing time as it is displayed in the UI

sale_order_line_position/models/sale_order.py
```diff
@@ -0,0 +1,28 @@
...
+    @api.depends("state")
+    def _compute_locked_positions(self):
+        for record in self:
+            record.locked_positions = record.state != "draft"
+
+    def action_quotation_send(self):
+        self.recompute_position()
+        return super().action_quotation_send()
+
+    def recompute_position(self):
```
sebalix:
```suggestion
    def recompute_positions(self):
```

sale_order_line_position/models/sale_order_line.py
```diff
@@ -0,0 +1,61 @@
...
+        if self.order_id.locked_positions:
+            return
+        lines = self.order_id.order_line.filtered(
+            lambda x: not x.display_type and x.sequence < self.sequence
+        )
+        self.position = len(lines) + 1
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        sale_ids = [
```
sebalix: Better to move all this override in its own method and call it from `create`.

sale_order_line_position/models/sale_order_line.py
```diff
@@ -0,0 +1,61 @@
...
+            line["order_id"]
+            for line in vals_list
+            if not line.get("display_type") and line.get("order_id")
+        ]
```
mmequignon: If we just create a section, then sale_ids will be empty.
Therefore, the following query will raise an exception:
```
[...]
  File "/odoo/external-src/sale-reporting/sale_order_line_position/models/sale_order_line.py", line 40, in create
    self.env.cr.execute(query, (ids,))
  File "/odoo/src/odoo/sql_db.py", line 173, in wrapper
    return f(self, *args, **kwargs)
  File "/odoo/src/odoo/sql_db.py", line 250, in execute
    res = self._obj.execute(query, params)
psycopg2.ProgrammingError: syntax error at or near ")"
LINE 3:         WHERE order_id in () GROUP BY order_id; 
```
I'll push a fix for that.

TDu: Thanks for your fix

sale_order_line_position/models/sale_order_line.py
TDu: Done, see last fixup

sale_order_line_position/models/sale_order.py
TDu: Changed

sale_order_line_position/models/sale_order_line.py
TDu: @simahawk This one has not been taken care of, should it ?

simahawk: if you can... ;)

simahawk:
pls squash once you've done

sale_order_line_position/models/sale_order_line.py
TDu: `can't we do this only on SO write of order_line?`
The write on the order is not called when a new line is created (At least not in the current test with create on a sale.order.line), And on sale creation neither.

`Also, does it make sense to do it for all order states?`
Yes because any sale.order.line should have a position and it should never change.

`I think this is adding too much overhead at lines' creation.`
Now with the `@api.model_create_multi` maybe not that much, on sale creation the `create` function is only called once!

TDu:
@simahawk I squashed, and answer on your previous comment about changing the implementation.

sale_order_line_position/models/sale_order_line.py
```diff
@@ -0,0 +1,68 @@
...
+        if self.order_id.locked_positions:
+            return
+        lines = self.order_id.order_line.filtered(
+            lambda x: not x.display_type and x.sequence < self.sequence
+        )
+        self.position = len(lines) + 1
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        vals_list = self._add_next_position_on_new_line(vals_list)
```
grindtildeath, line 29: Shouldn't you do a `copy` of `vals_list` to avoid any issue with mutable objects?

grindtildeath (approved):
Just a question but looks good :+1:

simahawk (approved):

sale_order_line_position/models/sale_order_line.py
simahawk: I don't think is necessary in this case.

PR #117, opened 2021-09-20
[13.0][IMP] sale_report_delivered_subtotal: Add delivered quantity on sale_report

CarlosRoca13:
cc @Tecnativa TT31881

please @sergio-teruel @chienandalu review this

### DIFF
--- a/sale_report_delivered_subtotal/reports/sale_report.py
+++ b/sale_report_delivered_subtotal/reports/sale_report.py
@@ -8,17 +8,27 @@ class SaleReport(models.Model):
     _inherit = "sale.report"
 
     price_subtotal_delivered = fields.Float(string="Subtotal Delivered")
+    weight_delivered = fields.Float("Gross Weight Delivered")
 
     def _query(self, with_clause="", fields=None, groupby="", from_clause=""):
         if fields is None:
             fields = {}
-        select_str = """ ,
+        select_str_price = """ ,
             sum((l.price_subtotal /
                  coalesce(nullif(l.product_uom_qty, 0), 1)
                 ) * l.qty_delivered)
             as price_subtotal_delivered
         """
-        fields.update({"price_subtotal_delivered": select_str})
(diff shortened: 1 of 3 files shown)


chienandalu (approved):

sergio-teruel (approved):

sale_report_delivered_subtotal/reports/sale_report.py
```diff
@@ -8,17 +8,27 @@ class SaleReport(models.Model):
...
+        select_str_price = """ ,
             sum((l.price_subtotal /
                  coalesce(nullif(l.product_uom_qty, 0), 1)
                 ) * l.qty_delivered)
             as price_subtotal_delivered
         """
-        fields.update({"price_subtotal_delivered": select_str})
+        select_str_weight = """ ,
+            sum(p.weight * l.qty_delivered /
+                u.factor * u2.factor) as weight_delivered
```
pedrobaeza, line 24: This is not quantity, so you shouldn't use UoM factors.

pedrobaeza (changes requested):

sale_report_delivered_subtotal/reports/sale_report.py
chienandalu: ~Right it's already converted:~ https://github.com/odoo/odoo/blob/13.0/addons/sale/report/sale_report.py#L69

Edit:

Sure, you're right (maybe a couple of parenthesis would give more clarity)

CarlosRoca13: I get this formula from this reference https://github.com/OCA/OCB/blob/25d0db569e26403777918ff179fe6d9993a694d5/addons/sale/report/sale_report.py#L95 changing the `product_uom_qty` by the `qty_delivered`

pedrobaeza (approved):
Yeah, sorry, it's alright.

PR #120, opened 2021-10-11
[13.0][FIX] sale-order-line-position

TDu:
The previous implementation with the onchange was broken and hardly
feasible (onchange are called on each line and the sequence set on them
changes on each call) and not possible to cover with unit
tests (reproduce the drag and drop).

The other solution could have been to replace the onchange with a
recompute of the position on every `write` of a sale.order but that
also leads to a complex solution.

As the positions are needed to discuss with the customer, the recompute of the
position on print, send and confirm is a good enough, simple solution.

- [ ] When reviewed it needs a forward port to 14.0

### DIFF
--- a/sale_order_line_position/models/ir_actions_report.py
+++ b/sale_order_line_position/models/ir_actions_report.py
@@ -1,0 +1,17 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import models
+
+
+class IrActionsReport(models.Model):
+    _inherit = "ir.actions.report"
+
+    def render_qweb_pdf(self, res_ids=None, data=None):
+        self.sale_recompute_positions(res_ids)
+        return super().render_qweb_pdf(res_ids, data)
+
+    def sale_recompute_positions(self, res_ids):
+        if self.model == "sale.order":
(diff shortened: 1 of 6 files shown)


TDu:
When reviewed it needs a forward port to 14.0

sale_order_line_position/models/ir_actions_report.py
```diff
@@ -0,0 +1,17 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import models
+
+
+class IrActionsReport(models.Model):
+    _inherit = "ir.actions.report"
+
+    def render_qweb_pdf(self, res_ids=None, data=None):
+        self.sale_recompute_position(res_ids)
```
rousseldenis: Is this really ? Launching code at report call ?

TDu: It does when printing from the UI

rousseldenis: My question is: Is it programmatically acceptable to do so ?

TDu: Ah, you mean a record should not risk being changed when printed ?
I can understand your point. But in this situation it is only information that is read-only that will be changed so I would think not that bad a practice.

PR #128, opened 2021-12-13
[14.0][MIG] sale_layout_category_hide_detail

rousseldenis:
Added:

- [x] Able to hide section line subtotal


rousseldenis:
@ernestotejeda This is an interesting widget.

But what do you think about making it more generic (only with the js part), in order to apply it on more apps that have form sections(purchase, ...) ?

ernestotejeda:
> But what do you think about making it more generic ...

I think that's a good idea. what do you think @pedrobaeza ?

pedrobaeza:
As the widget has been moved from sale to account during previous version (on the initial version of this module it wasn't), we can have an `account_layout_category_hide_detail` with account.move + widget stuff, and then add the sale one on top of it, as well as the purchase can be created easily.

I agree it can be a good move.

sale_layout_category_hide_detail/__manifest__.py
```diff
@@ -0,0 +1,24 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    "name": "Sale layout category hide detail",
+    "summary": "Hide details for sections in sale orders and invoices for "
+    "reports and customer portal",
+    "version": "14.0.1.0.0",
+    "category": "Sales Management",
+    "website": "https://github.com/OCA/sale-reporting",
+    "author": "Tecnativa, " "Odoo Community Association (OCA)",
```
ernestotejeda, line 11:
```suggestion
    "author": "Tecnativa, Odoo Community Association (OCA)",
```

ernestotejeda (commented):
Funcional review 👍🏼 

- I think readme should be improved to add explanation about 'Show line amount' option 
- You could take advantaje of the migration to adapt javascript code to ES, what do you think?

angelmoya (approved):

angelmoya (commented):
Functional review:

With "line amount hiden", pdf works, but website show details.

rousseldenis:
> Functional review:
> 
> With "line amount hiden", pdf works, but website show details.

Indeed. But after check, it was present in previous versions. I'll try to fix it

rousseldenis:
@angelmoya Could you try now ?

ernestotejeda (commented):
Code review: 👍🏼 
Functional review: 👍🏼 

I remember when I migrated the module to v12 @pedrobaeza suggested me to have the section name in the subtotal: https://github.com/OCA/sale-reporting/pull/70#pullrequestreview-259434580.
@pedrobaeza Is there a specific reason to do that or we  can keep the total without the section name as Rousseldenis has done now for v14?

pedrobaeza:
Yes, it's on purpose, and specified by our customer, as when having very big sections, you don't know which section you are putting the subtotal for. Can you please restore it, @rousseldenis, or add a configuration company parameter for specifying the behavior?

rousseldenis:
> Yes, it's on purpose, and specified by our customer, as when having very big sections, you don't know which section you are putting the subtotal for. Can you please restore it, @rousseldenis, or add a configuration company parameter for specifying the behavior?

Ok, try to do that. I understand the purpose. But IMHO the presentation is a little bit dirty. I'm not expert in UI. Keep in mind section name can be long and subtotal string can be 'lost in space'...

pedrobaeza:
One thing we can do is to put such text in a separate element for not moving `Subtotal` text.

elvise:
> One thing we can do is to put such text in a separate element for not moving `Subtotal` text.

Hi @rousseldenis great PR!!!
any plan for get it done ?

elvise:
@rousseldenis can we add an option for page break ?

angelmoya:
I try Todo page break in this version, and this is not easy to do, beacuse sections are inside table, and div page break do not work inside a table.

To do it, it is required to redefine table and create a table for section, and it's not possible to do inheriting views.

Maybe you can do other way, or maybe page break can work inside a table...

francesco-ooops:
Hi @rousseldenis , any plans for this one?

rousseldenis:
> Hi @rousseldenis , any plans for this one?

I'll try to get back here today or tomorrow

angelmoya:
Functional revision not OK (tested on runbot), on posted invoice clicking on button "Switch to: detail hiden" does visible the "paid" ribbon

image

rousseldenis:
> Functional revision not OK (tested on runbot), on posted invoice clicking on button "Switch to: detail hiden" does visible the "paid" ribbon
> 
> image

Thanks for the review but this module does not take care about that ribbon. Only lines.

ernestotejeda:
ping @rousseldenis :)

jeroenev:
> Thanks for the review but this module does not take care about that ribbon. Only lines.

If the module breaks default Odoo functionality, I think it's relevant to fix this before finishing this PR
because it does seem like that has an unintended impact on the payment banner

ao-landoo:
+1 to the bug, it is marking the invoice as paid when it shouldn't. I've been looking for the fix for a few hours but I cannot find the problem's solution...

sale_layout_category_hide_detail/static/src/js/boolean_fa_icon_widget.js
```diff
@@ -0,0 +1,82 @@
...
+        /**
+         * Toggle value
+         *
+         * @private
+         * @param {MouseEvent} event
+         */
+        _toggleValue: function (event) {
+            event.preventDefault();
+            event.stopPropagation();
+            this._setValue(!this.value);
```
ao-landoo, line 76: The invoice gets marked as paid when this line is executed

victoralmau: Reviewing this in detail would have a complex solution; however, most importantly, I believe that you should NOT be able to modify anything (use the buttons) if the invoice is not in draft. This change would therefore solve the problem. 
What do you think about it?

jeroenev: If that actually solves the problem then I think it's an acceptable solution

victoralmau:
Superseed by https://github.com/OCA/sale-reporting/pull/170

PR #132, opened 2022-02-07
15.0 mig sale layout category hide detail

jeroenev:
based on #128


jeroenev:
I've updated my email in the Readme to the one that was used on the signed CLA, seems like CLA-bot doesn't recognize the change?

rousseldenis:
> I've updated my email in the Readme to the one that was used on the signed CLA, seems like CLA-bot doesn't recognize the change?

See the message 'no github login found'. There is no more associated github account with that email.

If you have no more access to that email, I think you should re-sign the CLA.


FYI, I've added two commits in #128. One to allow hiding subtotal on section line, one to fix portal behaviour on hiding line amounts.

rousseldenis:
You can also review #128 in order to get it merged. Thanks

PR #133, opened 2022-02-10
[FIX][12.0] sale_comment_template: html field breaks boxed report

cubells:
Because html field always store `<p><br></p>` tags. So always is printed this table: https://github.com/OCA/sale-reporting/blob/12.0/sale_comment_template/views/report_saleorder.xml#L24

Before this patch:

before

After this patch:

after

### DIFF
--- a/sale_comment_template/views/report_saleorder.xml
+++ b/sale_comment_template/views/report_saleorder.xml
@@ -18,7 +18,7 @@
       </xpath>
 
       <xpath expr="//t[@t-foreach='doc.order_line']" position="inside">
-        <t t-if="line.formatted_note">
+        <t t-if="line.formatted_note != '&lt;p&gt;&lt;br&gt;&lt;/p&gt;'">
             <tr style="padding:0;">
                 <td colspan="7" style="padding:0;">
                     <table style="width:100%;border:0;padding:0;">


pedrobaeza:
Isn't there any standard method to check for such "empty" value instead of hardcoding it?

cubells:
I didn't found it.

I found that is used in several other code pieces:

https://github.com/OCA/OCB/blob/12.0/addons/website_forum/controllers/main.py#L343

https://github.com/OCA/knowledge/blob/12.0/document_page/models/document_page.py#L190

pedrobaeza:
OK, I see. Nasty patches...

And why is it codified instead of directly the raw HTML?

cubells:
I can change write method to avoid to be saved these tags, i.e.. If It is possible.

`And why is it codified instead of directly the raw HTML?`

I don't know. By changing the field type is another option.

cubells:
If we change `t-field `by `t-raw` nothing changes because the field contains the value `<p><br></p>`.

sale_comment_template/views/report_saleorder.xml
```diff
@@ -18,7 +18,7 @@
       </xpath>
 
       <xpath expr="//t[@t-foreach='doc.order_line']" position="inside">
-        <t t-if="line.formatted_note">
+        <t t-if="line.formatted_note != '&lt;p&gt;&lt;br&gt;&lt;/p&gt;'">
```
pedrobaeza, line 21: My question is why this is not:
```suggestion
        <t t-if="line.formatted_note != '<p><br></p>'">
```

pedrobaeza: OK, I have said nothing, it's an XML... Silly me

pedrobaeza (approved):

pedrobaeza:
Fast-tracking this:

/ocabot merge minor

PR #134, opened 2022-02-16
[13.0][IMP] sale_report_delivered: Add margin percent. Round amounts to avoid differences. Translations

carlosdauden:
@Tecnativa TT34568
cc @sergio-teruel @chienandalu

### DIFF
--- a/sale_report_delivered/reports/sale_report.py
+++ b/sale_report_delivered/reports/sale_report.py
@@ -3,7 +3,7 @@
 
 from psycopg2.extensions import AsIs
 
-from odoo import fields, models, tools
+from odoo import api, fields, models, tools
 
 
 class SaleReportDeliverd(models.Model):
@@ -21,7 +21,7 @@ class SaleReportDeliverd(models.Model):
     partner_id = fields.Many2one("res.partner", "Customer", readonly=True)
     company_id = fields.Many2one("res.company", "Company", readonly=True)
     user_id = fields.Many2one("res.users", "Salesperson", readonly=True)
-    price_subtotal = fields.Float("Untaxed Total", readonly=True)
+    price_subtotal = fields.Float("Untaxed total delivered", readonly=True)
     product_tmpl_id = fields.Many2one("product.template", "Product", readonly=True)
     categ_id = fields.Many2one("product.category", "Product Category", readonly=True)
     nbr = fields.Integer("# of Lines", readonly=True)
@@ -56,7 +56,8 @@ class SaleReportDeliverd(models.Model):
     order_id = fields.Many2one("sale.order", "Order", readonly=True)
     picking_id = fields.Many2one("stock.picking", "Picking", readonly=True)
     amount_cost = fields.Float("Amount cost", readonly=True)
-    margin = fields.Float("Margin", readonly=True)
+    margin = fields.Float("Margin delivered", readonly=True)
+    margin_percent = fields.Float(string="Margin delivered(%)", readonly=True)
 
     def _select(self):
         select_str = """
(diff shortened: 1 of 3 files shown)


carlosdauden:
ping @sergio-teruel @chienandalu

sale_report_delivered/reports/sale_report.py
```diff
@@ -56,7 +56,8 @@ class SaleReportDeliverd(models.Model):
     order_id = fields.Many2one("sale.order", "Order", readonly=True)
     picking_id = fields.Many2one("stock.picking", "Picking", readonly=True)
     amount_cost = fields.Float("Amount cost", readonly=True)
-    margin = fields.Float("Margin", readonly=True)
+    margin = fields.Float("Margin delivered", readonly=True)
+    margin_percent = fields.Float(string="Margin delivered(%)", readonly=True)
```
chienandalu, line 60: Aren't percentage values quite useless (and missleading) in a report? If I get a 50% for a product in 10 orders, my margin would be 500% as shown in the report :upside_down_face:

sergio-teruel (approved):

sale_report_delivered/reports/sale_report.py
carlosdauden: The percentage is calculated on the fly depending on the displayed values.
Inicialiced as zero:
https://github.com/OCA/sale-reporting/pull/134/files#diff-0ea6a8e310c8aedd7d64cd2e9dc00f983471b36f5254176023547b7f678bc2eeR99
Computed here:
https://github.com/OCA/sale-reporting/pull/134/files#diff-0ea6a8e310c8aedd7d64cd2e9dc00f983471b36f5254176023547b7f678bc2eeR243

chienandalu: Oh, nice! :+1:

chienandalu (approved):

sale_report_delivered/reports/sale_report.py
```diff
@@ -90,10 +91,15 @@ def _select(self):
...
+                ROUND(sub.unsigned_purchase_price * unsigned_product_uom_qty,
+                      {decimal_places}))) AS amount_cost,
             sum(signed_qty * unsigned_price_subtotal - COALESCE(-sub.amount_cost, signed_qty *
-                sub.unsigned_purchase_price * unsigned_product_uom_qty)) AS margin
-        """
+                ROUND(sub.unsigned_purchase_price * unsigned_product_uom_qty,
+                      {decimal_places}))) AS margin,
+            0 AS margin_percent
+        """.format(
+            decimal_places=self.env.company.currency_id.decimal_places
```
pedrobaeza: This should use the currency of the sales order, not the company currency.

carlosdauden: I was about to put a fixed 2...
This is a way to put an appropriate value without complicating the query and penalizing performance.
When someone opens an issue I promise to contemplate it. :wink:

pedrobaeza: Carlos, the query is not much complicated. Just JOIN res_currency and the decimal places are there. This is not correct and may lead to weird data that is very difficult to debug. Please change it.

carlosdauden: I don't know if you are not understanding the context or if you're too maniacal :stuck_out_tongue_winking_eye: , but I don't want to penalize even a single millisecond for a query that can already be quite heavy.
Selección_524

If you are afraid of "weird data that is very difficult to debug", approve this PR because it is precisely what it is solving. :smile:

pedrobaeza: OK, I'll approve it if you put a comment `FIXME:` in the code about this question.

carlosdauden: Changes done.
Making the changes costs me less than writing the FIXME and that we are always with the debate :sweat_smile:

pedrobaeza:
Thanks:

/ocabot merge patch

PR #135, opened 2022-02-21
[13.0][IMP] sale_report_delivered: Discard customer returned stock moves that not to refund

sergio-teruel:
cc @Tecnativa TT34713

ping @carlosdauden @chienandalu

### DIFF
--- a/sale_report_delivered/reports/sale_report.py
+++ b/sale_report_delivered/reports/sale_report.py
@@ -164,8 +164,15 @@ class SaleReportDeliverd(models.Model):
         return from_str
 
     def _where(self):
+        """Take into account only stock moves from internal locations to other
+        locations and moves from customer with the field 'to_refund' True
+        """
         return """
-            WHERE sm.state = 'done' OR sm.state IS NULL
+            WHERE (sm.state = 'done' OR sm.state IS NULL) AND (
+                (source_location.usage = 'internal' AND dest_location.usage = 'customer') OR
+                (source_location.usage = 'customer' AND dest_location.usage = 'internal'
+                    AND sm.to_refund)
+            )
         """
 
     def _group_by(self):


sale_report_delivered/reports/sale_report.py
```diff
@@ -165,7 +165,7 @@ def _from(self):
 
     def _where(self):
         return """
-            WHERE sm.state = 'done' OR sm.state IS NULL
+            WHERE sm.to_refund = false AND (sm.state = 'done' OR sm.state IS NULL)
```
pedrobaeza: For shorter form, but shouldn't be the other way (to include those to refund)?

```suggestion
            WHERE NOT sm.to_refund AND (sm.state = 'done' OR sm.state IS NULL)
```

sergio-teruel:
@pedrobaeza review again please!!

sale_report_delivered/reports/sale_report.py
```diff
@@ -165,7 +165,9 @@ def _from(self):
 
     def _where(self):
         return """
-            WHERE sm.state = 'done' OR sm.state IS NULL
+            WHERE (sm.state = 'done' OR sm.state IS NULL) AND
+                  (source_location.usage <> 'customer' AND
```
pedrobaeza: Aren't you discarding with this all returns?

chienandalu: Yes, maybe an easier fix could be to set the `signed_qty` to 0 for this special case

sergio-teruel: I think it is better understood this way

carlosdauden (approved):

pedrobaeza (approved):

sale_report_delivered/reports/sale_report.py
```diff
@@ -164,8 +164,15 @@ def _from(self):
...
     def _where(self):
+        """Take into account only stock moves from internal locations to other
+        locations and moves from customer with the field 'to_refund' True
+        """
         return """
-            WHERE sm.state = 'done' OR sm.state IS NULL
+            WHERE (sm.state = 'done' OR sm.state IS NULL) AND (
+                (source_location.usage = 'internal' AND dest_location.usage = 'customer') OR
+                (source_location.usage = 'customer' AND dest_location.usage = 'internal'
+                    AND sm.to_refund)
```
chienandalu, line 174: This leaves out dropshipping, isn't it?

carlosdauden: You're right.
A new PR could be made by changing both `usage = 'internal'` to `usage IN ('internal', 'supplier')`

PR #141, opened 2022-04-21
[15.0] add base_multicompany_reporting_currency

yankinmax:
This module is needed to set Group currency that will be applied to all companies.
group_curr


rousseldenis:
@yankinmax I don't see where that parameter is applied to all companies. Do I miss something ?

base_group_currency/models/res_config_settings.py
```diff
@@ -0,0 +1,14 @@
...
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    company_group_currency_id = fields.Many2one(
+        "res.currency",
+        string="Group Currency",
+        config_parameter="base_group_currency.group_currency_id",
```
yankinmax: @rousseldenis here is it

base_group_currency/data/default_group_currency.xml
```diff
@@ -0,0 +1,6 @@
+<odoo>
+    <record id="group_currency_id" model="ir.config_parameter">
+        <field name="key">base_group_currency.group_currency_id</field>
+        <field name="value" ref="base.EUR" />
```
yankinmax: @rousseldenis and this is default value

rousseldenis:
Ok, what I wanted to say is following your description, we could think that when creating a new company, the default currency will be that one.

yankinmax:
@rousseldenis I'm not good at writing descriptions)))
How would you rephrase it to be more clear? Pls, I really appreciate your help.

rousseldenis:
> @rousseldenis I'm not good at writing descriptions))) How would you rephrase it to be more clear? Pls, I really appreciate your help.

I will try but not too much time.

First, you need to identify the functional needs for this. For what will it be used?

base_group_currency/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Base Group Currency",
+    "summary": "Adds the possibility to specify Group currency",
+    "version": "15.0.1.0.1",
```
i-vyshnevska:
```suggestion
    "version": "15.0.1.0.0",
```

i-vyshnevska (approved):

i-vyshnevska:
@rousseldenis  this currency field will be used in other places to compare amounts in different companies and documents, by standalone this addon doesn't bring functionality only in dependency with other development

rousseldenis:
> @rousseldenis this currency field will be used in other places to compare amounts in different companies and documents, by standalone this addon doesn't bring functionality only in dependency with other development

This should be added in README and IMHO, in configuration you need to add a help too. I think the 'Group' term is maybe inappropriate. Maybe 'Inter-Company'

base_group_currency/__manifest__.py
yankinmax: bad copy-paste, thanks

ajaniszewska-dev (approved):

gurneyalex (approved):

rousseldenis (approved):

PR #142, opened 2022-04-21
[15.0] add crm group currency

yankinmax:
This Module adds Amount Group Currency to CRM Lead.


rousseldenis:
@yankinmax Thanks for this. But should it be here instead of crm repo ?

yankinmax:
@rousseldenis the idea is to have it here, because Total Group Currency field will be used for reporting purposes. It makes Lead's list more informative.
And I'll also add `sale_group_currency` similar to `crm_group_currency`, so to have it one place is useful to maintain and improve it in future.
BTW thank you for review, could you pls help me to solve last issue? IDK why it argues on `test-requiremnts.txt`.

rousseldenis:
> IDK why it argues on `test-requiremnts.txt`.

That's in order to be sure to remove that line before merging

yankinmax:
@rousseldenis ah ok, now I see the meaning) THX

rousseldenis:
> @rousseldenis the idea is to have it here, because Total Group Currency field will be used for reporting purposes. It makes Lead's list more informative

I understand but in fact you add fields in crm leads that could be used elsewhere. I think the base data should stay on its own repo and the report itself here.

yankinmax:
> > @rousseldenis the idea is to have it here, because Total Group Currency field will be used for reporting purposes. It makes Lead's list more informative
> 
> I understand but in fact you add fields in crm leads that could be used elsewhere. I think the base data should stay on its own repo and the report itself here.

Then I propose to leave here `base_group_currency` and `sale_group_currency`, `crm_group_currency` will move to OCA/crm.
Is it ok for you?

rousseldenis:
I think you should a little bit explain more what is 'Group' currency

yankinmax:
ok, I'll add more details to README

yankinmax:
@rousseldenis suppressed by https://github.com/OCA/crm/pull/426

yankinmax:
> ok, I'll add more details to README

updated README in `base_group_currency` https://github.com/OCA/sale-reporting/pull/141

PR #149, opened 2022-05-06
[13.0][ADD] sale_report_delivered_volume: New module to show the delivered volume at sale.report

CarlosRoca13:
cc @Tecnativa TT36336

please @sergio-teruel @carlosdauden review it


sale_report_delivered_volume/models/sale_report.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class SaleReport(models.Model):
+    _inherit = "sale.report"
+
+    volume_delivered = fields.Float(digits="Volume")
+
+    # pylint:disable=dangerous-default-value
+    def _query(
+        self, with_clause="", fields={}, groupby="", from_clause=""  # noqa: B006
```
sergio-teruel:
```suggestion
        self, with_clause="", fields=None, groupby="", from_clause=""
```

sergio-teruel (changes requested):

sergio-teruel:
Can you add the OCA icon ??
icon

CarlosRoca13:
@sergio-teruel done!

sale_report_delivered_volume/models/sale_report.py
```diff
@@ -0,0 +1,23 @@
...
+    _inherit = "sale.report"
+
+    volume_delivered = fields.Float(digits="Volume")
+
+    def _query(self, with_clause="", fields=None, groupby="", from_clause=""):
+        if fields is None:
+            fields = {}
+        fields[
+            "volume_delivered"
+        ] = ", sum(p.volume * l.qty_delivered / u.factor * u2.factor) as volume_delivered"
```
carlosdauden, line 17: Why extend this method instead of extend _select 
https://github.com/OCA/sale-reporting/blob/caa9682f871866ba042c503b756c700c5cd77961/sale_report_delivered/reports/sale_report.py#L62
or/and _sub_select
https://github.com/OCA/sale-reporting/blob/caa9682f871866ba042c503b756c700c5cd77961/sale_report_delivered/reports/sale_report.py#L103

carlosdauden (changes requested):

sale_report_delivered_volume/models/sale_report.py
CarlosRoca13: We are extending `sale`, not `sale_report_delivered`

carlosdauden: Sorry... The module name...

carlosdauden (approved):

CarlosRoca13:
Probably we should change it to `sale_report_volume_delivered`, what do you think @carlosdauden

pedrobaeza:
But why don't rely in `sale_report_delivered` for providing the quantity and only multiply by the unit volume?

PR #151, opened 2022-05-11
[13.0][FIX] sale_report_delivered: The weight and volume of returns is increasing the total instead of decreasing it

CarlosRoca13:
cc @Tecnativa TT36635

please @carlosdauden @sergio-teruel review this


sale_report_delivered/reports/sale_report.py
```diff
@@ -140,8 +140,16 @@ def _sub_select(self):
...
+            p.weight * (
+                CASE
+                WHEN spt.code = 'incoming' THEN -1
+                ELSE 1 END
+            ) * sm.product_uom_qty / u.factor * u2.factor as weight,
+            p.volume * (
+                CASE
+                WHEN spt.code = 'incoming' THEN -1
+                ELSE 1 END
+            ) * sm.product_uom_qty / u.factor * u2.factor as volume,
```
carlosdauden: Instead of this, you can replace lines 87 and 88
https://github.com/OCA/sale-reporting/blob/8752ecb7cd1cfc9c1731384fb1bf86937161dbcd/sale_report_delivered/reports/sale_report.py#L87-L88
by:
            sum(**signed_qty** * sub.weight) as weight,
            sum(**signed_qty** * sub.volume) as volume,

carlosdauden (changes requested):

carlosdauden (approved):

sergio-teruel (approved):

PR #153, opened 2022-05-12
[15.0][MIG] sale_order_line_position

ajaniszewska-dev:


simahawk (commented):
LG but please rewrite your commits to include the module name

sebalix (approved):

simahawk (approved):

simahawk:
Sorry, just noticed the 1st commit before the mig is kind of useless by itself, it should be squashed on top of the migration one.
Not a blocker tho.

PR #170, opened 2022-08-26
[14.0][MIG] sale_layout_category_hide_detail: Migration to 14.0

victoralmau:
Superseed https://github.com/OCA/sale-reporting/pull/128

Changes done:
- [x] Migration to 14.0
- [x] Do not allow to use the buttons depending on the status of the order or invoice (https://github.com/OCA/sale-reporting/pull/128#issuecomment-1180554504).
- [x] Set a fixed width for the button columns. 
- [x] Simplify code in functions to allow inheritability.

Please @pedrobaeza can you review it?

@Tecnativa TT38579


francesco-ooops (commented):
@victoralmau is there any way to improve the layout? The three buttons are taking up to 25% of the sale order line width

image

especially in SOs without sections: 

image

victoralmau:
> @victoralmau is there any way to improve the layout? The three buttons are taking up to 25% of the sale order line width

Although this in v13 is similar, I have applied the change to optimize space as much as possible.

francesco-ooops:
> > @victoralmau is there any way to improve the layout? The three buttons are taking up to 25% of the sale order line width
> 
> Although this in v13 is similar, I have applied the change to optimize space as much as possible.

is there a specific reason for this fourth empty column? (sorry, I don't know the module very well and I don't see it in the past versions)

image

victoralmau:
This column must be shown because depending on the corresponding use case, the corresponding icon will be shown or not.

francesco-ooops (approved):
Functional review ok!

sale_layout_category_hide_detail/static/src/js/boolean_fa_icon_widget.js
```diff
@@ -36,28 +37,46 @@ odoo.define("sale_layout_category_hide_detail.boolean_fa_icon_widget", function
...
-            var terminology = this.attrs.options.terminology;
-            var hover_true =
-                (terminology && _t(terminology.hover_true)) || _t("Click to uncheck");
-            var hover_false =
-                (terminology && _t(terminology.hover_false)) || _t("Click to check");
-            var tip = this.value ? hover_true : hover_false;
-            // Set template and add it to $el
-            var template = "<span class='fa %s' title='%s' aria-label='%s'></span>";
-            this.$el.empty().append(_.str.sprintf(template, fa_class, tip));
+            if (this.recordData.display_type === "line_section") {
```
pedrobaeza: Why do you need this `if` if this is only applied for this widget?

sale_layout_category_hide_detail/views/sale_order_report_templates.xml
```diff
@@ -52,7 +52,7 @@
             expr="//tbody[hasclass('sale_tbody')]//tr[hasclass('is-subtotal', 'text-right')]/td/strong"
             position="after"
         >
-      <strong class="mr16">Subtotal <span t-field="current_section.name" /></strong>
+      <strong class="mr16">Subtotal</strong>
```
pedrobaeza: This is ripping a feature that our client needs, so please restore it

pedrobaeza: You can remove the commit.

elvise (approved):
LGTM!

sale_layout_category_hide_detail/views/sale_order_report_templates.xml
victoralmau: Ok, commit removed.

sale_layout_category_hide_detail/static/src/js/boolean_fa_icon_widget.js
victoralmau: Actually it is iterating through all the lines (not only section lines), and for example at debug level it is a bit confusing; anyway I have removed it because in any case it would be something that should be implemented by odoo core.

francesco-ooops:
@pedrobaeza good now?

francesco-ooops:
@victoralmau could you fix urls to icons in description files? 

@pedrobaeza this is an issue I've seen happening almost everytime there is an image/screenshot in description file, eg [here](https://github.com/OCA/purchase-workflow/tree/14.0/purchase_quick), dunno if that's mentioned in guidelines or can be prevented differently

jeroenev (commented):
EDIT: line not correctly selected, see review below

sale_layout_category_hide_detail/static/src/js/boolean_fa_icon_widget.js
```diff
@@ -0,0 +1,98 @@
...
+        // --------------------------------------------------------------------------
+
+        /**
+         * Toggle value
+         *
+         * @private
+         * @param {MouseEvent} event
+         */
+        _toggleValue: function (event) {
+            this._allowEdit();
```
jeroenev, line 89: don't think this line does anything? only the last line seems to have an impact here

victoralmau: https://github.com/OCA/sale-reporting/pull/172

PR #178, opened 2022-10-17
[15.0][ADD]sale_blanket_order_line_position

leemannd:
- move base methods about the position computation to a mixin. 
- add the module sale_blanket_order_position

The goal is to have the possibility to use the position on order/blanket lines.


StephaneMangin (approved):

sale_blanket_order_line_position/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+    "version": "15.0.1.0.0",
+    "category": "Sales",
+    "author": "Camptocamp, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "website": "https://github.com/OCA/sale-reporting",
+    "depends": ["sale_blanket_order", "sale_order_line_position"],
+    "data": [
+        "views/blanket_order_views.xml",
+        # "views/res_config_settings.xml",
+        # "report/sale_order_report.xml",
```
leemannd: Missing into the PR

sale_blanket_order_line_position/tests/test_blanket_line_position.py
```diff
@@ -0,0 +1,162 @@
...
+                    "product_id": self.product.id,
+                    "product_uom": self.product.uom_id.id,
+                    "original_uom_qty": 15.0,
+                },
+            ]
+        )
+        self.assertEqual(new_line.position, 4)
+        self.blanket_order.state = "open"
+        self.blanket_order.line_ids[0].unlink()
+        self.assertEqual(new_line.position, 3)
```
leemannd: This is wrong

cyrilmanuel (approved):
LG for me, interesting mixin !

sale_order_line_position/models/position_mixin.py
```diff
@@ -0,0 +1,62 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl)
+
+from odoo import api, fields, models
+
+
+class OrderPositionMixin(models.AbstractModel):
```
mmequignon, line 7: please, create one file per class:
https://www.odoo.com/documentation/15.0/contributing/development/coding_guidelines.html#file-naming

sale_blanket_order_line_position/models/__init__.py
```diff
@@ -0,0 +1,3 @@
+from . import blanket_orders
```
mmequignon, line 1: please, no plurals in file names (or module names)

sale_blanket_order_line_position/models/blanket_order_line.py
```diff
@@ -0,0 +1,35 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import models
+
+
+class BlanketOrderLine(models.Model):
+    _name = "sale.blanket.order.line"
+    _inherit = ["sale.blanket.order.line", "order.line.position.mixin"]
+
+    def _add_next_position_on_new_line(self, vals_list):
```
mmequignon, line 11: looks like this code is almost exactly the same than the one on `sale.order`.
Could it be factored?
You could use `self._name` and `self._table` perhaps.

sale_blanket_order_line_position/models/blanket_orders.py
```diff
@@ -0,0 +1,21 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import models
+
+
+class BlanketOrder(models.Model):
+    _name = "sale.blanket.order"
+    _inherit = ["sale.blanket.order", "order.position.mixin"]
+
+    def recompute_positions(self):
```
mmequignon, line 11: same thing here.
I feel like it could be done in the mixin.

leemannd:
Hello @OCA/crm-sales-marketing-maintainers A review and eventually a merge are welcome

sale_blanket_order_line_position/models/blanket_orders.py
mmequignon: Also, I would have added an `@api.depends("")` here, and dropped the `ir.action.report` override.
```python
@api.depends(
    "locked_positions",
    "company_id.disable_sale_position_recompute",
    "line_ids.display_type",
    "line_ids.sequence",
)
```

sale_order_line_position/models/position_mixin.py
```diff
@@ -0,0 +1,62 @@
...
+    _name = "order.position.mixin"
+    _description = "Order position mixin"
```
mmequignon, line 9: You can add a class attribute like `position_field_name`, which would have to be set as `order_lines` on `sale.order` and so on.
By doing that, you just have to define this attribute in order to set that up, instead of having to define methods in classes extending this mixin.

mmequignon: you can take this as an example:
https://github.com/OCA/stock-logistics-warehouse/blob/14.0/stock_packaging_calculator/models/product_qty_by_packaging_mixin.py#L14

sale_order_line_position/models/position_mixin.py
```diff
@@ -0,0 +1,62 @@
...
+
+    locked_positions = fields.Boolean(compute="_compute_locked_positions")
+
+    @api.depends("state")
```
grindtildeath, line 13: Here you're depending on a field that is not part of the mixin :confused:

sale_order_line_position/models/position_mixin.py
```diff
@@ -0,0 +1,62 @@
...
+        for record in self:
+            record.position_formatted = record._format_position(record.position)
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        vals_list = self._add_next_position_on_new_line(vals_list)
+        return super().create(vals_list)
+
+    def unlink(self):
+        sales = self.mapped("order_id")
```
grindtildeath, line 48: Here you're using a field that is not part of the mixin neither

oca_dependencies.txt
```diff
@@ -0,0 +1,2 @@
+# See https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#oca_dependencies-txt
```
simahawk, line 1: you can trash this file

sale_blanket_order_line_position/models/blanket_orders.py
simahawk: It can stay on on the mixin if you pass the lines.
The `depends` won't work. Each model has its own field... "order_line", "line_ids", etc.
Also, is more reliable if we call it when/where we need it.

sale_blanket_order_line_position/models/ir_actions_report.py
```diff
@@ -0,0 +1,17 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import models
+
+
+class IrActionsReport(models.Model):
+    _inherit = "ir.actions.report"
+
+    def render_qweb_pdf(self, res_ids=None, data=None):
+        self.blanket_recompute_positions(res_ids)
```
simahawk, line 11: Why do you need this? The order of the lines must be always the correct one before printing.

sale_blanket_order_line_position/tests/test_blanket_line_position.py
```diff
@@ -0,0 +1,162 @@
...
+from datetime import date, timedelta
+
+from odoo import fields
+from odoo.tests import TransactionCase
+
+
+class TestBlanketOrderLinePosition(TransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
```
simahawk, line 12: tracking_disable pls

sale_order_line_position/models/position_mixin.py
simahawk, line 13:
```suggestion
    @api.depends(self._compute_locked_positions_depends())
```
and let the consumer of the mixin pass the fields

sale_order_line_position/models/position_mixin.py
```diff
@@ -0,0 +1,62 @@
...
+        res = super().unlink()
+        for sale in sales:
+            sale.recompute_positions()
```
simahawk, line 51:
```suggestion
        self.recompute_positions()
    
    def recompute_positions(self):
        raise NotImplementedError()
```

sale_order_line_position/models/position_mixin.py
```diff
@@ -0,0 +1,62 @@
...
+    def action_confirm(self):
+        self.recompute_positions()
+        return super().action_confirm()
+
+    def action_quotation_send(self):
+        self.recompute_positions()
+        return super().action_quotation_send()
+
+    def recompute_positions(self):
+        return NotImplementedError
```
simahawk, line 27:
```suggestion
    def recompute_positions(self):
        if not self.locked_positions:
            self._recompute_positions()
    
    def _recompute_positions(self):
        return NotImplementedError
```

See https://github.com/OCA/sale-reporting/pull/195

simahawk (changes requested):

PR #179, opened 2022-10-21
12.0 add sale report wizard

victor-champonnois:


sale_report_wizard/wizard/sale_report_wizard.xml
```diff
@@ -0,0 +1,28 @@
...
+    <field name="arch" type="xml">
+        <form>
+        <group>
+        <field name="report" />
+        <field name="filter_date_type" />
+        <field
+                        name="start_date"
+                        attrs="{'invisible': [('filter_date_type', '=', False)]}"
+                    />
+        <field
```
robinkeunen: strange formatting

sale_report_wizard/wizard/sale_report_wizard.py
```diff
@@ -0,0 +1,61 @@
...
+        view_id = self.env.ref(self.report.view).id
+        search_view_id = ""
+        if self.report.search_view:
+            search_view_id = self.env.ref(self.report.search_view).id
+
+        domain = []
+
+        # sale.report name date_order as date,
+        # while report.all.channel.sale and sale.order 
+        # name it as date_order
```
robinkeunen, line 35: Is there other things that make report.all.channel.sale incompatible? Otherwise, your comment in DESCRIPTION.rst is outdated.

robinkeunen (approved):
Very interesting. Small comments.

robinkeunen (changes requested):
You're missing CONTRIBUTORS.rst though, otherwise 👍

sale_report_wizard/wizard/sale_report_wizard.xml
victor-champonnois: fixed

sale_report_wizard/wizard/sale_report_wizard.py
victor-champonnois: Ah, yes. It lacked an overwrite of the report.all.channel.sale query to work. It was in the custom module rotordc_sale_reports_top_category. 
So:
- I also added it here.
- Removed the comment in Description.

victor-champonnois:
@robinkeunen thx for the review, contributors.rst added

robinkeunen (approved):

sale_report_wizard/report/report_all_channels_sales.py
```diff
@@ -0,0 +1,73 @@
...
+
+from odoo import fields, models
+
+
+class PosSaleReport(models.Model):
+    _inherit = "report.all.channels.sales"
+
+    # Allow sale_report_wizard module to use the report.all.channels.sale 
+    # model by adding confirmation_date to it.
+    confirmation_date = fields.Datetime(string="Confirmation Date", readonly=True)
```
DucTruongKomit, line 11: You don't need set the string here, because Odoo will display the field as "Confirmation Date"

sale_report_wizard/report/report_all_channels_sales.py
```diff
@@ -0,0 +1,73 @@
...
+    date = fields.Datetime(string="Confirmation Date", readonly=True)
```
DucTruongKomit, line 12: There are 2 issues here for the field name:
1. The name of field is the convention in python => I suggest we should change this field to something else, for ex: so_confirmation_date = (string='Sale Confirmation Date').
2. The string of 2 fields are duplicated, so it will cause the log of system have more warning and also display 2 labels the same will make users confused.

DucTruongKomit:
Hello please have a look on my comment @victor-champonnois

victor-champonnois:
> Hello please have a look on my comment @victor-champonnois

@DucTruongKomit I don't see a comment from you in this PR. What comment are you talking about ?

DucTruongKomit:
Sorry, I totally forgot submit my reviewing

sale_report_wizard/report/report_all_channels_sales.py
DucTruongKomit: Or maybe this one is duplicated?

DucTruongKomit:
Also, there are something about Checks, you need to take care of.
image

PR #191, opened 2022-11-24
[13.0][FIX] product_sold_by_delivery_week: Fix key error when product not in dict

carlosdauden:
@Tecnativa TT40635

### DIFF
--- a/product_sold_by_delivery_week/models/sale_order.py
+++ b/product_sold_by_delivery_week/models/sale_order.py
@@ -40,5 +40,5 @@ class SaleOrderLine(models.Model):
         for line in to_process_lines:
             partner = line.get_partner_for_reporting()
             line.weekly_sold_delivered_shown = _format_weekly_string(
-                partner_products_weekly[partner][line.product_id]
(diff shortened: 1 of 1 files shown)


product_sold_by_delivery_week/models/sale_order.py
```diff
@@ -40,5 +40,5 @@ def _compute_weekly_sold_delivered_shown(self):
         for line in to_process_lines:
             partner = line.get_partner_for_reporting()
             line.weekly_sold_delivered_shown = _format_weekly_string(
-                partner_products_weekly[partner][line.product_id]
+                partner_products_weekly[partner].get(line.product_id, False)
```
chienandalu: Does it need to be  `False` or could it be `None`?

carlosdauden: Changed

chienandalu (approved):

sergio-teruel (approved):

CarlosRoca13 (approved):

PR #199, opened 2022-12-19
[14.0][ADD] sale_order_terms_document

remi-filament:
Add a module allowing to add Terms and Conditions in PDF after the generated sale order PDF.
Also, this module adds a link to T&C PDF in portal view when signing document


pedrobaeza:
Please don't add plurals in module names. Anyway, there's already `sale_comment_template` for this.

remi-filament:
Hi @pedrobaeza,
Thanks for looking into it !

About plural, here I refer to Terms & Conditions, which is why I used plural, not sure if it still makes sense to call it term in this context ?

Also, from my point of view, use case is different from `sale_comment_template` one. For the following reasons :

1. Here it is proposed to directly load a PDF file which will be added after sale order (this PDF with Terms and Conditions you probably already have if you are an existing sales company and you do not want to rewrite it fully in Odoo HTML field)
2. Terms and Conditions make no sense on invoices and therefore should not be present on invoice
3. You need to make these terms and conditions visible / accessible on portal view so that when customer signs sale order, he acknowledges T&C (this modules allows for this, when sale_comment_template does not)

With the above explanation, do you still think this module is useless (since already covered in another module) ?
Best Regards,

pedrobaeza:
I think you can integrate the PDF possibility on `sale_comment_template` (well, more on `base_comment_template`), and also a check about propagating or not to invoice.

remi-filament:
Yes probably, though it adds a lot of complexity for customers when most of them do not need it...

pedrobaeza:
I don't think there's too much burden to add this in the other module, and it's very related, but it's up to you to do it. If you continue this way, then the module should be handled as something more generic, as this isn't only for sales orders. It should in OCA/reporting-engine as well. The name may be `report_extra_pdf_content` or similar, and includes a conditional expression to be safe-evaled to include it or not.

njeudy:
Pour moi c'est ok d'ajouter ce module tel quel:

1. parce que ca répond à un besoin client (plusieurs client meme)
2. parce qu'il fonctionne

For me it's ok to add this module as is:

1. because it meets a customer need (several customers even)
2. because it works

I think that we can make this module general if the need is expressed by several people and that it will not pose any problem thereafter?

@pedrobaeza , I would like to understand what are the guidelines or fears that make you want to make it more generic ?

A little module that allows you to add the T&C on the quotes, sounds good to me :)

will review this soon thanks @remi-filament

pedrobaeza:
My goal is to avoid a user fragmentation and several things for the same goal but different implementations that bloats an instance. It's legit that you have some terms that are written, and another that includes a PDF attached, but I'm not enforcing anything, I'm just giving my opinion. And it's a pity that the feature of having extra PDF content to any report is limited in visibility and applicability in this repo.

It's also very counter-intuitive that for making this module works, you have to go to the report definition and enabling an option that can be enabled in other models reports.

sale_order_terms_document/controllers/portal.py
```diff
@@ -0,0 +1,26 @@
...
+
+from odoo.addons.portal.controllers.portal import CustomerPortal
+
+
+class CustomerPortal(CustomerPortal):
+    @http.route(
+        ["/my/orders/tcs/<int:tc_id>"], type="http", auth="public", website=True
+    )
+    def download_tc(self, tc_id):
+        tc_id = request.env["document.tc"].sudo().browse(tc_id)
```
DucTruongKomit: This is the object not ID, so I think we should use the variable:
document_tc = request.env["document.tc"].sudo().browse(tc_id)

sale_order_terms_document/models/document_tc.py
```diff
@@ -0,0 +1,13 @@
...
+from odoo import fields, models
+
+
+class DocumentTc(models.Model):
+    _name = "document.tc"
+    _description = "Terms and Conditions Documents"
+
+    document = fields.Binary("Attached Document")
+    name = fields.Char("Document Name")
+    active = fields.Boolean("Active", default=True)
```
DucTruongKomit: You don't need the "Active" string here, unless if you want to display the label of fields as different name.

remi-filament:
I agree with @pedrobaeza that having to configure the boolean in ir.actions.report is confusing. I just removed it from the code, so that adding terms and conditions PDF will only happen on sale.action_report_saleorder.
Also, I agree that it could be made more general elsewhere, but we do not have another use case where it makes sense elsewhere and therefore I prefer to keep things simple to answer the very common need from our customers to be able to add terms and conditions on their sale orders without having to configure anything else or add many dependencies that add complexity on my instances.

sale_order_terms_document/models/document_tc.py
remi-filament: Thanks I removed it !

PR #209, opened 2023-01-10
[15.0][FIX] sale_multicompany_reporting_currency: Fix multicompany_reporting_currency_rate calculation

yankinmax:


ajaniszewska-dev (approved):

ivantodorovich (approved):
Could be good to add unit tests, as codecov mentions the case where `record.company_id` is not set is not tested

yankinmax:
@ivantodorovich I would say I've written `_compute_multicompany_reporting_currency_rate` as much as close to `compute_currency_rate` to avoid any issues:
https://github.com/odoo/odoo/blob/b396584515b51035c14f100bf6b90fc9f7606284/addons/sale/models/sale_order.py#L308
But, the condition when `if not order.company_id` seems redundant to me, because `company_id` is required on db level
https://github.com/odoo/odoo/blob/b396584515b51035c14f100bf6b90fc9f7606284/addons/sale/models/sale_order.py#L259
So, no need to write extra test here.

yankinmax:
@gurneyalex can you pls help to get it merged?

sale_multicompany_reporting_currency/models/sale_order.py
```diff
@@ -93,7 +98,6 @@ def _compute_amount_multicompany_reporting_currency(self):
             ):
                 to_amount = reporting_amount
             else:
-                to_amount = (
-                    reporting_amount / record.multicompany_reporting_currency_rate
-                )
+                inverse_rate = 1.0 / record.multicompany_reporting_currency_rate
+                to_amount = reporting_amount / inverse_rate
```
SilvioC2C: Mathematically, to divide by the inverse rate is equivalent to multiplying by the original rate, so this snippet of code is equivalent to:
```suggestion
                to_amount = reporting_amount * record.multicompany_reporting_currency_rate
```
which (I believe) is easier to read

SilvioC2C (approved):
Small suggestion, else LGTM

yankinmax:
NOTE: pre-commit is failing because of an error caused by pre-commit initialisation itself. So, it's not related to a module.

simahawk:
@yankinmax can you pls remove the odoo version from the commit msg?

simahawk:
also... pre-commit is failing

simahawk:
Need an update https://github.com/OCA/oca-addons-repo-template/issues/185
https://github.com/OCA/sale-reporting/pull/213

simahawk:
FTR pre-commit was failing due to a broken download likely. Reported here anyway https://github.com/OCA/oca-addons-repo-template/issues/186

PR #222, opened 2023-06-02
[16.0][MIG] sale_layout_category_hide_detail: Migration to version 16.0

Yadier-Tecnativa:
@Tecnativa TT43243


pedrobaeza (changes requested):
See the image:

Selección_030

Yadier-Tecnativa:
> See the image:
> 
> Selección_030

Thanks for the review.
For the first problem, I'll have to check what action trigger the create method.
and the second, the width is because there are 2 more hidden columns in this state

pedrobaeza:
I have tried again and now the icons are not clickable. Please ping when all of this is solved.

Yadier-Tecnativa:
ping @victoralmau , @pedrobaeza

pedrobaeza:
Almost there. Please move the subtotal when you hide the details to the right:

Selección_043


And please reduce even more the width of each icon column to the minimum needed.

Yadier-Tecnativa:
Done with css styles

pedrobaeza:
The width of each icon column is still too narrow.

sale_layout_category_hide_detail/views/invoice_report_templates.xml
```diff
@@ -33,7 +33,7 @@
             />
     </xpath>
     <xpath
-            expr="//tbody[hasclass('invoice_tbody')]//tr[hasclass('is-subtotal', 'text-right')]"
+            expr="//tbody[hasclass('invoice_tbody')]//tr[hasclass('is-subtotal', 'text-end')]"
```
pedrobaeza, line 36: Why `right` is not working?

pedrobaeza (commented):
What happens with translations of the tooltip?

sale_layout_category_hide_detail/views/invoice_report_templates.xml
Yadier-Tecnativa: that class was changed in the base template sale.report_saleorder_document:
```
 <tr class="is-subtotal text-end">

Comment (Yadier-Tecnativa) [reply sale_layout_category_hide_detail/views/invoice_report_templates.xml:36]: Updated missing translation

Comment (pedrobaeza) [reply sale_layout_category_hide_detail/views/invoice_report_templates.xml:36]: This is still not working:

Selección_057

**Comment by Yadier-Tecnativa** [comment]
thanks for the review @pedrobaeza ; 
Added the styles in assets for the 4 different views,(sale orders and invoices) for reports and customer portal

**Comment by Yadier-Tecnativa** [comment]
ping @pedrobaeza , @CarlosRoca13

**Review by CarlosRoca13 — COMMENTED** [review]
It's possible to avoid doble clicking on icon to achieve the result?

Something like boolean_toggle does

On next gif you can see what I mean:
Example_pr

**Comment by Yadier-Tecnativa** [comment]
> It's possible to avoid doble clicking on icon to achieve the result?

The rendered widget in section_note needs to have focus to be able to change its value because in this case we added readonly and AllowEdition properties for orders with state != 'draft' which might be colliding with edit mode and Onchange events directly from the Checkout.
This condition is logged for improvement in future releases.

File: sale_layout_category_hide_detail/static/src/xml/boolean_fa_icon_widget.xml [hunk 2]
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<templates xml:space="preserve">
+
+    <t
+        t-name="sale_layout_category.BooleanFaIconWidget"
+        t-inherit="web.BooleanField"
+        t-inherit-mode="primary"
```
CarlosRoca13:
```suggestion
```

sale_layout_category_hide_detail/static/src/xml/boolean_fa_icon_widget.xml
```diff
@@ -0,0 +1,22 @@
...
+            <attribute name="className">'fa-icon-widget'</attribute>
+        </xpath>
+        <xpath expr="//CheckBox" position="inside">
+            &#8203; <!-- Zero width space needed to set height -->
+            <div class="fa_icon_tooltip top" role="tooltip">
+                <span class="tip_text" t-out="state.text_tooltip" />
+            </div>
+            <span t-attf-class="fa #{state.fa_class}" />
+        </xpath>
+    </t>
```
CarlosRoca13:
```suggestion
        <div t-on-click="onClickButton">
            <div class="fa_icon_tooltip top" role="tooltip">
                <span class="tip_text" t-out="state.text_tooltip" />
            </div>
            <span t-attf-class="fa #{state.fa_class}" />
        </div>
      </t>
```

sale_layout_category_hide_detail/static/src/js/boolean_fa_icon_widget.js
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    onChange(newValue) {
+        if (this.props.readonly || !this.isAllowEdit()) {
+            return;
+        }
+        super.onChange(newValue);
+        this.state.fa_class = this.getFaIconClass(newValue);
+        this.state.text_tooltip = this.getTextTooltip(newValue);
+    }
```
CarlosRoca13:
```suggestion
    onClickButton(ev) {
        ev.stopPropagation();
        if (this.props.readonly || !this.isAllowEdit()) {
            return;
        }
        const newValue = !this.props.value
        this.props.update(newValue);
        this.state.fa_class = this.getFaIconClass(newValue);
        this.state.text_tooltip = this.getTextTooltip(newValue);
    }
```

Yadier-Tecnativa: Thanks for the tips @CarlosRoca13 ,the issue was solved.

pedrobaeza:
Almost there. The preview of the SO for section with detail hidden is not showing the subtotal:

Selección_075

http://oca-sale-reporting-16-0-pr222-d80e0d12f3ad.runboat.odoo-community.org/web#model=sale.order&id=25&action=377&view_type=form&cids=1

On invoices is working correctly.

Yadier-Tecnativa:
Fixed up

sale_layout_category_hide_detail/static/src/js/sale_layout_category_hide_detail.js
```diff
@@ -1,86 +1,52 @@
...
-                ) {
-                    $th.text("").removeClass("o_column_sortable");
-                }
-                return $th;
-            },
-        };
+patch(SectionAndNoteListRenderer.prototype, "new_widgets_buttons_patch", {
+    // Define the patched method here
+    setup() {
+        this._super.apply(this, arguments);
```
chienandalu: ?

sale_layout_category_hide_detail/static/src/js/sale_layout_category_hide_detail.js
```diff
@@ -1,86 +1,52 @@
...
+    },
 
-        sectionAndNoteListRenderer.include(SectionAndNoteListRenderer);
-    }
-);
+    // Override
+    getSectionColumns(columns) {
+        const sectionCols = columns.filter(
```
chienandalu: No the way to override. Call super and add your condition

sale_layout_category_hide_detail/static/src/js/sale_layout_category_hide_detail.js
chienandalu:
```suggestion
        super(...arguments);
```

sale_layout_category_hide_detail/static/src/js/sale_layout_category_hide_detail.js
```diff
@@ -1,86 +1,52 @@
-/* eslint-disable init-declarations */
-/* Copyright 2019 Tecnativa - Ernesto Tejeda
-/* Copyright 2022 Tecnativa - Víctor Martínez
- * License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
```
chienandalu: Keep credits and license

sale_layout_category_hide_detail/static/src/js/boolean_fa_icon_widget.js
```diff
@@ -1,101 +1,108 @@
...
+        var fa_icons = this.props.fa_icons;
+        var icon_true = fa_icons.icon_true || iconTrue;
+        var icon_false = fa_icons.icon_false || iconFalse;
+        return currentValue ? icon_true : icon_false;
+    }
 
-            // --------------------------------------------------------------------------
-            // Public
-            // --------------------------------------------------------------------------
+    getTextTooltip(currentValue) {
```
chienandalu: Please document your coude with proper jsdoc comments: https://github.com/odoo/odoo/wiki/Javascript-coding-guidelines#documentation

sale_layout_category_hide_detail/static/src/css/boolean_fa_icon_widget.scss
```diff
@@ -0,0 +1,35 @@
+div.fa-icon-widget > input.form-check-input,
```
chienandalu: Why not using bootstrap classes in the component template to avoid as much css as possible?

sale_layout_category_hide_detail/static/src/css/invoice_report_style.scss
```diff
@@ -0,0 +1,3 @@
+span.float-right {
```
chienandalu: This looks like a patch to something else having side effects :thinking:

chienandalu (changes requested):

CarlosRoca13 (approved):

sale_layout_category_hide_detail/static/src/css/invoice_report_style.scss
Yadier-Tecnativa: Thanks for the review, I reduced the css to 0 this time updating to the bootstrap new classes.

Yadier-Tecnativa:
Done!

pedrobaeza (approved):
OK at functional level.

@chienandalu do you agree at JS/CSS level?

.eslintrc.yml
chienandalu, line 1: Did you change this manually?

sale_layout_category_hide_detail/static/src/js/boolean_fa_icon_widget.js
```diff
@@ -1,101 +1,113 @@
...
+            state !== "draft"
+        ) {
+            this.props.allow = false;
+        }
+        return this.props.allow;
+    }
+
+    isReadonly() {
+        return this.props.record.isReadonly(this.props.name);
+    }
```
chienandalu: This is a getter the inherited class: https://github.com/odoo/odoo/blob/ce3e66609e70cef6801baf9dba71c84798d382c1/addons/web/static/src/views/fields/boolean/boolean_field.js#L11-L13

Anyway, it would be nice to have documentation on why this is redefined

sale_layout_category_hide_detail/static/src/js/boolean_fa_icon_widget.js
```diff
@@ -0,0 +1,113 @@
...
+
+    getFaIconClass(currentValue) {
+        var fa_icons = this.props.fa_icons;
+        var icon_true = fa_icons.icon_true || iconTrue;
+        var icon_false = fa_icons.icon_false || iconFalse;
+        return currentValue ? icon_true : icon_false;
+    }
+
+    getTextTooltip(currentValue) {
```
chienandalu: Why not using a getter to simplify everythin related with this? Take a look at this reference: https://github.com/odoo/odoo/blob/ce3e66609e70cef6801baf9dba71c84798d382c1/addons/web/static/src/views/fields/field.js#L186

chienandalu (commented):
Overall I'm missing a minimum documentation on the js widget code. Without a deep analysis I think it could be simplified as well. What do you think @CarlosRoca13 ?

.eslintrc.yml
CarlosRoca13: Now I understand why precommit wasn't throwing an error about esm modules...

Please @Yadier-Tecnativa discard this changes and change the extension `*.js` files to `*.esm.js`

chienandalu: The old trick :laughing:

sale_layout_category_hide_detail/static/src/js/boolean_fa_icon_widget.js
Yadier-Tecnativa: Ok, good suggestion

.eslintrc.yml
Yadier-Tecnativa: Extensions have been changed

sale_layout_category_hide_detail/static/src/js/boolean_fa_icon_widget.js
Yadier-Tecnativa: Documented in the code

sale_layout_category_hide_detail/static/src/js/boolean_fa_icon_widget.js
Yadier-Tecnativa: Nice example, but it doesn't work at the moment, I tried before and the behavior was not as expected.
This property is as if it were computed, it depends on the onchange event of this.props.value.
I even used an asynchronous event, but the fa_icons and text_tooltip state always fails the first time it's clicked.
Please if you can take a look and propose a better solution it will be welcome.

chienandalu (approved):
Thanks @Yadier-Tecnativa :)

PR #226, opened 2023-08-02
[16.0][MIG] Migration of module sale_comment_template

alan196:
### Proposed changes

Migration PR for module sale_comment_template.

### Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Migration Update

### Checklist

- [x] I have read the CONTRIBUTING doc
- [x] I have signed the CLA
- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the necessary documentation


alan196:
@pedrobaeza @EduardoJarsa Could you help me to review this PR 🙏

EduardoJarsa (commented):
functional review, everything looks working

EduardoJarsa (approved):
Functional review, everything looks working

sale_comment_template/views/base_comment_template_view.xml
```diff
@@ -5,11 +5,8 @@
         <field name="type">ir.actions.act_window</field>
         <field name="res_model">base.comment.template</field>
         <field name="view_mode">tree,form</field>
-        <field name="domain">[('model_ids.model', '=', 'sale.order')]</field>
-        <field
-            name="context"
-            eval="{'default_model_ids': [(4, ref('sale.model_sale_order'))]}"
-        />
+        <field name="domain">[('models', 'like', 'sale.order')]</field>
```
pedrobaeza: This is not correct, as if you have a template for `sale.order.line`, it will be shown.

alan196: But it will fail if you share this template with account.move
This template couldn't be shared with another model

What do you propose? @pedrobaeza

pedrobaeza: Use the proper domain. There is another m2m field if I don't remember bad for filtering directly against `ir.model`, and using `ref`, you can get what is needed.

alan196: Thanks for your help

I have tested with 

```xml
<field name="domain" eval="[('model_ids.id', '=', ref('sale.model_sale_order'))]" />
```
```xml
<field name="domain" eval="[('model_ids, '=', ref('sale.model_sale_order'))]" />
```
```xml
<field name="domain">[('model_ids.model', '=', 'sale.order')]</field>
```

None of this works, the action returns an empty page

Any idea how can I implement this domain correctly?

alan196: Another option is to not filter at all... 

https://github.com/OCA/account-invoice-reporting/blob/16.0/account_comment_template/views/base_comment_template_view.xml

Only show the menuitem without any domain

pedrobaeza: Search method seems to not be well implemented:

https://github.com/OCA/reporting-engine/blob/16.0/base_comment_template/models/base_comment_template.py#L119

alan196:
@pedrobaeza I updated the action in order to work with the defined search method.

Now it works, Could you help me to review 🙏

pedrobaeza (approved):
/ocabot merge nobump

PR #227, opened 2023-08-21
[16.0][IMP] sale_comment_template: allow to use dynamic information in template

alan196:
With this change, it allows to define dynamic information from the sale order in the template

### DIFF
--- a/sale_comment_template/views/report_saleorder.xml
+++ b/sale_comment_template/views/report_saleorder.xml
@@ -10,7 +10,7 @@
                 t-foreach="doc.comment_template_ids.filtered(lambda x: x.position == 'before_lines')"
                 t-as="comment_template_top"
             >
-                <div t-esc="comment_template_top.text" />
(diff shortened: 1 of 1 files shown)


pedrobaeza (changes requested):
- `t-raw` is deprecated. Use `t-out`. Anyway, that's not for dynamic expressions, but for HTML.
- This is not `[ADD]`, but `[IMP]`.

alan196:
@pedrobaeza thanks for the review 👍 

I have updated the requested changes

pedrobaeza (approved):

alan196:
@pedrobaeza I just updated the PR to use qweb rendering instead of inline rendering engine, or do you think is better to make a PR to base_comment_template module in order to make qweb rendering the default?

pedrobaeza:
I think that the syntax should be defined at template level.

YahairaCVJarsa (approved):

PR #235, opened 2023-10-17
[15.0][ADD] sale_section_and_note_revamp

StephaneMangin:
This module provides high level method helpers to ease sale sections and notes management and use elsewhere.


sale_section_and_note_revamp/__init__.py
```diff
@@ -0,0 +1 @@
+from . import models
```
Ricardoalso, line 1: Missing tests import

sale_section_and_note_revamp/models/display_line_mixin.py
```diff
@@ -0,0 +1,154 @@
...
+        :param order_line: a sale.order.line record
+        """
+        self.ensure_one()
+        raise NotImplementedError
+
+    def inject_sections_and_notes(self):
+        """This method inject all related display lines to the right position
+        for the inheriting model
+
+        See sale.order::compute_order_lines_dependency for further explanations
```
leemannd: `calc_order_lines_dependencies` ?

leemannd (approved):

sale_section_and_note_revamp/models/display_line_mixin.py
```diff
@@ -0,0 +1,177 @@
...
+    def get_section(self):
+        """Returns the section of the current line"""
+        if not self.is_section():
+            previous_record = self.previous_line_id
+            while previous_record:
+                if previous_record.is_section():
+                    return previous_record
+                previous_record = previous_record.previous_line_id
+
+    def get_note(self):
```
leemannd, line 61: The method get_section loops in previous records to find the section.

I find it counter-intuitive that this method only checks if the next line is a note. I would expect this method to loop in previous records to find and get the related note, or a different naming

sale_section_and_note_revamp/tests/common.py
```diff
@@ -0,0 +1,106 @@
...
+                "partner_id": cls.partner_a.id,
+                "partner_invoice_id": cls.partner_a.id,
+                "partner_shipping_id": cls.partner_a.id,
+                "pricelist_id": cls.company_data["default_pricelist"].id,
+            }
+        )
+
+        cls.sol_section_1 = SaleOrderLine.create(
+            {
+                "sequence": 1,
```
leemannd, line 29: Has odoo will popluate all lines with sequence=10, wouldn't it be better to not force the sequence at creation and change it only if required in tests?

leemannd:
Hello @OCA/crm-sales-marketing-maintainers a review and eventually a merge are welcome

sale_section_and_note_revamp/models/display_line_mixin.py
```diff
@@ -0,0 +1,154 @@
...
+
+    def has_section(self):
+        """Quick method helper"""
+        return bool(self.get_section())
+
+    def has_note(self):
+        """Quick method helper"""
+        return bool(self.get_note())
+
+    def get_section(self):
```
mmequignon, line 52: I'm pretty sure we can avoid having references to other SO lines by using simple logic, as we do have a sequence we can use.
This would avoid (maybe) lots of ORM calls.
```python3
line_ids = self.order_id.line_ids.ids # [1, 2 (section), 3, 4 (self), 5, 6, 7] (get them sorted by sequence)
line_ids = line_ids.reverse() # [7, 6, 5, 4, 3, 2, 1]
line_index = line_ids.index(self.id) # 3
above_ids = line_ids[3:] # [4 (self), 3, 2 (section), 1] (maybe we can safely use line_ids[3+1:] as we're sure current line isn't a section)
above_lines = self.browse(above_ids)
for line in above_lines:
     if line.is_section(): # true for line with id 2
          return line
```

sale_section_and_note_revamp/models/display_line_mixin.py
mmequignon: I don't get it either.

sale_section_and_note_revamp/models/display_line_mixin.py
```diff
@@ -0,0 +1,154 @@
...
+        if self.is_section():
+            result = self.env[self._name]
+            next_record = self.next_line_id
+            while next_record:
+                if next_record.is_section():
+                    break
+                result |= next_record
+                next_record = next_record.next_line_id
+            notes = result.filtered(lambda r: r.is_note())
+            return result if with_notes else result - notes
```
mmequignon, line 95: Same here, no need of those `next_line_id` and `previous_line_id`
```python3
        if self.is_section():
            res = self if with_notes else self.browse()
            line_ids = self.order_id.line_ids.ids
            line_index = line_ids.index(self.id)
            next_line_ids = line_ids[line_index+1:]
            next_lines = self.browse(next_line_ids)
            for line in next_lines:
                 if line.is_section():
                     break
                  res |= line
            return res
```

mmequignon (commented):
I'm not very sure about this.
I'm not fond of using the orm recursively.

rousseldenis:
@StephaneMangin What's the status of this ?

simahawk:
@leemannd can you take over?

PR #237, opened 2023-10-19
[15.0][IMP] sale_report_delivered_subtotal: Use price_reduce field when price_subtotal is zero

carlosdauden:
@Tecnativa
Ping @sergio-teruel @CarlosRoca13

### DIFF
--- a/sale_report_delivered_subtotal/reports/sale_report.py
+++ b/sale_report_delivered_subtotal/reports/sale_report.py
@@ -14,10 +14,12 @@ class SaleReport(models.Model):
         if fields is None:
             fields = {}
         select_str_price = """ ,
-            sum((l.price_subtotal /
-                 coalesce(nullif(l.product_uom_qty, 0), 1)
-                ) * l.qty_delivered)
-            as price_subtotal_delivered
+        sum(
+            CASE WHEN l.price_subtotal <> 0
+                THEN (l.price_subtotal / l.product_uom_qty)
+                ELSE l.price_reduce
+            END
+            * l.qty_delivered) as price_subtotal_delivered
         """
         select_str_weight = """ ,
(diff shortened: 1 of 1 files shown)


sergio-teruel (approved):

pedrobaeza:
`price_reduce` mustn't be used, because it contains the unit price already rounded to the price digits, provoking rounding errors. Talking with Odoo sales dev in latest Odoo XP, they see that this field should be removed for avoiding confusions and I think it has been done in latest versions.

carlosdauden:
Changes made to use price_reduce only when price cannot be calculated from price_subtotal.

sale_report_delivered_subtotal/reports/sale_report.py
```diff
@@ -14,10 +14,12 @@ def _query(self, with_clause="", fields=None, groupby="", from_clause=""):
...
             fields = {}
         select_str_price = """ ,
-            sum((l.price_subtotal /
-                 coalesce(nullif(l.product_uom_qty, 0), 1)
-                ) * l.qty_delivered)
-            as price_subtotal_delivered
+        sum(
+            CASE WHEN l.price_subtotal <> 0
+                THEN (l.price_subtotal / l.product_uom_qty)
+                ELSE l.price_reduce
```
pedrobaeza, line 20: Why not doing here the operation of price_unit - discount for getting a price reduced without rounding?

carlosdauden: Because for these rare cases I prefer the result to be more compatible with other modules than more accurate.

pedrobaeza (approved):
OK, good enough

/ocabot merge patch

pedrobaeza:
Please fw-port it to v16.

PR #243, opened 2023-12-13
[14.0][ADD] sale_report_pagebreak

petrus-v:
Control page breaks in PDF sale report.

This split the main table by section and avoid
as much as possible breaking page inside a section.


petrus-v:
waiting https://github.com/OCA/sale-reporting/pull/245 to fix pre-commit job

Issue #246, opened 2023-12-20
[16.0]Sale_comment_template: not render html in report


odooNextev:
![image](https://github.com/OCA/sale-reporting/assets/93990941/b77235bc-10d2-4c58-88f4-71a7ad5e78d5)

PR #247, opened 2023-12-20
[IMP]sale_comment_template: render html in report

odooNextev:
#246


odooNextev:
@pedrobaeza small fix but important.
What do you opinion? Thanks

pedrobaeza:
`t-raw` is deprecated. You should consider using `Markup` instead, but this fix is not correct.

odooNextev:
Thanks @pedrobaeza, I will open a new pr on reporting-engine to add this improvement to base_comment_template

odooNextev:
This is the pr to fix it https://github.com/OCA/reporting-engine/pull/833

PR #251, opened 2024-02-13
[15.0][FIX] sale_report_salesman: import hooks init file

carolinafernandez-tecnativa:
- Add on init file to import hooks file

@Tecnativa

@pedrobaeza


pedrobaeza (approved):
Fast-tracking the obvious patch

/ocabot merge patch

PR #252, opened 2024-02-16
[FIX] product_sold_by_delivery_week: work with lower permissions

yajo:
A user with basic stock and sale permissions couldn't validate a picking because it meant writing a field into the product. Now that write is issued with `sudo()`.

I refactored a bit the tests. They were testing very weird scenarios that would never happen in the real world. And, what's more interesting, they were not testing the real world scenarios.

I lowered the permissions in tests, to make sure that the module works with the basic permissions.

Finally, I also changed some spelling mistakes that I simply could not help but change.

@moduon MT-5114


product_sold_by_delivery_week/README.rst
```diff
@@ -57,7 +57,7 @@ info in other views like product recommendations, this is not needed.)
 Usage
 =====
 
-Once configured, hints will apperar as an optional column in sale order
+Once configured, hints will aperar as an optional column in sale order
```
chienandalu: :thinking: I see I'm not the only typo blind in the house :smile: 
```suggestion
Once configured, hints will appear as an optional column in sale order
```

Or maybe we could just use *show up*

chienandalu (commented):
Thanks @yajo :wink: 

Tiny remark:

product_sold_by_delivery_week/README.rst
yajo: haah ooops

Shide (approved):
Code review LGTM

chienandalu (approved):

EmilioPascual (approved):

yajo:
Thanks everyone.

/ocabot merge patch

PR #257, opened 2024-03-25
[16.0] sale_order_line_position + sale_layout_category_hide_detail fix

simahawk:
Fix `sale_order_line.py:38: DeprecationWarning: Deprecated method flush(), use flush_model(), flush_recordset() or env.flush_all() instead`

Collateral fix for `sale_layout_category_hide_detail` for odoo code change.

PR #258, opened 2024-03-26
[16.0][FIX] sale_order_line_position: remove deprecated function

trisdoan:


santostelmo (approved):

cyrilmanuel (approved):

sale_order_line_position/models/sale_order_line.py
```diff
@@ -35,7 +35,7 @@ def _add_next_position_on_new_line(self, vals_list):
         ]
         if sale_ids:
             ids = tuple(set(sale_ids))
-            self.flush_model()
+            self.env.flush_all()
```
simahawk, line 38: `flush_model` is not deprecated and in fact I fixed the old `flush` call here https://github.com/OCA/sale-reporting/pull/257.

Plus, why flushin ALL the models instead of just the ones that you need? In this case just sale_order_line.

simahawk (changes requested):
Am I missing anything?

sale_order_line_position/models/sale_order_line.py
trisdoan: Hello @simahawk, indeed, the warning no longer exists. I recall it was there :thinking: 

Anyway, I closed this PR

PR #259, opened 2024-03-26
[16.0][ADD] sale_report_tax_display

victorvermot:
This module modifies the invoices and sales order reports.
- It adds options to customize both reports for each res.partner
- It allows you to hide the tax lines on invoices and sales order.
- It also allows you to show the amount on both reports with or without taxes included.
- The columns name of the reports are modified accordingly depending on the options selected.


cyrilmanuel (approved):
lgtm

PR #260, opened 2024-04-02
[15.0][FIX] sale_comment_template: fix action to allow to see templates in sales

GasparJarsa:
This commit addresses the following issues:
Resolved an issue where users were unable to view templates in the comment list due to an incorrect domain configuration in the action.
Updated the context to set the sale order as the default option.

### DIFF
--- a/sale_comment_template/views/base_comment_template_view.xml
+++ b/sale_comment_template/views/base_comment_template_view.xml
@@ -5,11 +5,8 @@
         <field name="type">ir.actions.act_window</field>
         <field name="res_model">base.comment.template</field>
         <field name="view_mode">tree,form</field>
-        <field name="domain">[('model_ids.model', '=', 'sale.order')]</field>
-        <field
-            name="context"
-            eval="{'default_model_ids': [(4, ref('sale.model_sale_order'))]}"
-        />
+        <field name="domain">[('model_ids', '=', 'sale.order')]</field>
+        <field name="context">{'default_models': 'sale.order'}</field>
         <field
             name="view_id"
             ref="base_comment_template.view_base_comment_template_tree"


GasparJarsa:
@alan196 @SirAionTech Could you please help me checking this PR?

alan196 (approved):
Technical and functional review 👍

sale_comment_template/views/base_comment_template_view.xml
```diff
@@ -5,11 +5,8 @@
         <field name="type">ir.actions.act_window</field>
         <field name="res_model">base.comment.template</field>
         <field name="view_mode">tree,form</field>
-        <field name="domain">[('model_ids.model', '=', 'sale.order')]</field>
-        <field
-            name="context"
-            eval="{'default_model_ids': [(4, ref('sale.model_sale_order'))]}"
-        />
+        <field name="domain">[('models', '=', 'sale.order')]</field>
```
SirAionTech: This would only find templates that are _only_ for `sale.order`s.

According to https://github.com/OCA/purchase-reporting/commit/84a78d6c648fad422746f8eb7e504e2f60f2bd08, this
```suggestion
        <field name="domain">[('model_ids', '=', 'sale.order')]</field>
```
should allow to find `sale.order`'s templates when they are listed along with other models' templates (as it should have been before this change), would you mind including it?
In the commit you also have an example of how to implement a test case, it would be great if you could include that too.

sale_comment_template/views/base_comment_template_view.xml
```diff
@@ -5,11 +5,8 @@
...
+        <field name="context" eval="{'default_models': 'sale.order'}" />
```
SirAionTech:
```suggestion
        <field name="context">{'default_models': 'sale.order'}</field>
```
`eval` was needed for `eval`uating `ref`; it works either way so this is not blocking.

SirAionTech (changes requested):
Thanks for the PR!

alan196:
@SirAionTech Thanks for your review! I apply your suggestion and rebase the code, could you help me to review 🙏

SirAionTech (approved):

AaronHForgeFlow (approved):
Functional :+1:

AaronHForgeFlow:
I think this can be merged

AaronHForgeFlow:
Can someone reopen this? I have been using this for a while now, and IMHO it should be included.

PR #261, opened 2024-04-04
[16.0][MIG] base_multicompany_reporting_currency: Migration to 16.0 

peluko00:
Module migrated to version 16.0

cc https://github.com/APSL 153695

@miquelalzanillas @lbarry-apsl @javierobcn @mpascuall please review

Issue https://github.com/OCA/sale-reporting/issues/174

PR #262, opened 2024-04-15
[16.0][MIG] sale_multicompany_reporting_currency: Migration to 16.0 

vvrossem:
Depends on https://github.com/OCA/sale-reporting/pull/261


yankinmax:
For OCA PRs depending on another PR: the recommended way is to use test-requirements.txt.
There you can specify your python dependency as:
`odoo${10|11|12|13|14}-addon-${addon_name} @ git+https://github.com/OCA/$repo_name@refs/pull/$PR_NR/head#subdirectory=setup/$addon_name`
For this to work, you MUST have setup folder updated in the base PR.
Normally this is done on pre-commit run on recent versions BUT it works only when the module is installable. Hence, if you are migrating a module, make sure the setup folder is up to date and staged in the changes.
If you want to generate it manually, just run `setuptools-odoo-make-default -d .` in the repo folder.
Note that from odoo 15 we don't need the version in the python pkg name anymore.
When you use such approach, always add or update test-requirements.txt in a separated commit to make prominent that there's a commit to trash before merging.

strating from v17
`odoo-addon-${addon_name} @ git+https://github.com/OCA/$repo_name.git@refs/pull/$PR_NR/head#subdirectory=${addon_name}`

test-requirements.txt
```diff
@@ -1 +1 @@
-odoo-addon-base_multicompany_reporting_currency @ git+https://github.com/OCA/sale-reporting.git@refs/pull/261/head#subdirectory=base_multicompany_reporting_currency
+odoo16-addon-base_multicompany_reporting_currency @ git+https://github.com/OCA/sale-reporting.git@refs/pull/261/head#subdirectory=base_multicompany_reporting_currency
```
yankinmax:
```suggestion
odoo-addon-base_multicompany_reporting_currency @ git+https://github.com/OCA/sale-reporting@refs/pull/261/head#subdirectory=setup/base_multicompany_reporting_currency
```

yankinmax: if doesn't work for you try this

yankinmax (approved):

sale_multicompany_reporting_currency/models/res_config_settings.py
```diff
@@ -0,0 +1,21 @@
...
+    def set_values(self):
+        sale_order = self.env["sale.order"]
+        applied_currency = sale_order._get_multicompany_reporting_currency_id()
+        super().set_values()
+        to_apply_currency = self.multicompany_reporting_currency
+        if applied_currency.id != to_apply_currency.id:
+            sale_order.search([]).write(
+                {"multicompany_reporting_currency_id": to_apply_currency.id}
+            )
+        return True
```
imlopes, line 21: @vvrossem thanks for your work ;)

IMO we need to update the ``amount_multicompany_reporting_currency`` field on the SO's every time we change the settings, to have the right displayed amount.

```suggestion
    def set_values(self):
        sale_order = self.env["sale.order"]
        applied_currency = sale_order._get_multicompany_reporting_currency_id()
        super().set_values()
        to_apply_currency = self.multicompany_reporting_currency
        sales = sale_order.search([])
        if applied_currency.id != to_apply_currency.id:
            sales.write(
                {"multicompany_reporting_currency_id": to_apply_currency.id}
            )
        else:
            sales._compute_amount_multicompany_reporting_currency()
        return True
```

vvrossem: @imlopes thank you for noticing this behavior. It's fixed: https://github.com/OCA/sale-reporting/pull/262/commits/84b1bf73329e4394bf388cc73ca1069274d212e5

yankinmax: do we really need a dependency on `company_id.amount_option`?
When you change a system setting `set_values` is called and normally with 
```
sales.write({"multicompany_reporting_currency_id": to_apply_currency.id})
```
existing orders with different currencies will be recomputed (I mean `multicompany_reporting_currency_id` and `amount_multicompany_reporting_currency`).
I don't think we need either `sales._compute_amount_multicompany_reporting_currency()`, because it's additional recomputation on saving settings. But, redundant as we recompute **orders in EURO to orders in EURO**. Why do we need that additional recompute which will impact the performance?
On large databases the settings saving can take some time

vvrossem: Without `@api.depends("company_id.amount_option", (...)`, `amount_multicompany_reporting_currency` is not recomputed when changing `amount_option` in the system settings

PR #268, opened 2024-05-15
[14.0][FIX] sale_layout_category_hide_detail: Full width to note lines 

victoralmau:
Note type lines do not need to add any logic, they need to maintain full width.

**Before**
antes

**After**
despues

Please @pedrobaeza and @CarlosRoca13 can you review it?

@Tecnativa TT49131


pedrobaeza:
`renderBodyCell` is not saying anything. Can you please explain a bit more in the commit message the goal? And doing this is overloading each render with one extra RPC. You should do it on JS on my opinion depending on the display_type.

victoralmau:
Changed to another approach (updated commit description with before and after examples).

pedrobaeza (approved):

chienandalu (approved):

PR #269, opened 2024-05-24
[15.0][FIX] sale_layout_category_hide_detail: Full width to note lines

victoralmau:
FWP from 14.0: https://github.com/OCA/sale-reporting/pull/268

Note type lines do not need to add any logic, they need to maintain full width.

**Before**
antes

**After**
despues

Please @pedrobaeza and @CarlosRoca13 can you review it?

@Tecnativa TT49131


pedrobaeza (approved):
Fast-tracking as already merged in previous version:

/ocabot merge patch

PR #284, opened 2024-09-30
[16.0][FIX] sale_order_line_position: use formated field in report and make…

ng-ife:
… sure it is always the first element


pedrobaeza (approved):
Let's fast-track it:

/ocabot merge minor

PR #286, opened 2024-10-01
[17.0] sale_order_line_position: position numbers first and use formatted field in so

ng-ife:
* make sure position number always come first on reports
* replace outdated bootstrap class


gbrito (approved):

Kimkhoi3010 (approved):
LGTM

PR #289, opened 2024-10-23
[FIX] [REF] [ADD] [16.0] sale_report_delivered_deposit

Shide:
First of all, sorry for all this stuff in one PR.
It's necessary to make some refactor to make it work.

There are 3 commits:
1. 47da083e5f5a2ddf7262922174c083f6cfb9fcde [FIX]: Remove the COALESCE to let the NULL value go to the SELECT (it's in the sub-select) and let use the product_cost of the sale order line
2. df08f64a0c777c31ea9a2670f58a95c3ee94f233 [REF]: Refactor to split the conditions in separate functions to allow a better inheritance in the `sale_report_delivered_deposit` module and other future modules, allowing them to add its own casuistic.
3. 4694cf7e9f24fe39d6d6fa42c357f2fe71b20e6f [ADD]: The new module to show Customer Deposits on the Sale Report Delivered.

MT-7240 @moduon @rafaelbn @yajo @sergio-teruel please review if you want :)


Gelojr:
Hi @Shide 
When an order is placed to generate a delivery from the customer's deposit, the cost of the product is calculated in the sales order line. The module stock_customer_deposit has already managed that the cost in the sales order in which the delivery will be made is 0, because this cost is already taken into account in the order in which the deposit will be made. In this way we duplicate it and both in the deposit delivery order and in the report there is a negative margin.
**SO deposit**
image


**SO Deposit deliver**
image

Report
image

I know this PR affects the deposit slips that show up on the report, but could it be affecting the cost of the line for some reason?

Shide:
> I know this PR affects the deposit slips that show up on the report, but could it be affecting the cost of the line for some reason?

For that case, you have this module: https://github.com/OCA/stock-logistics-workflow/tree/16.0/stock_customer_deposit_sale_margin

When it's a deposit, marks purchase_price = 0.0, so the margin remains 0.0%

Shide:
Also reviewing `stock_customer_deposit_sale_margin` i found a bug: https://github.com/OCA/stock-logistics-workflow/pull/1749

Gelojr (approved):
@Shide With the clarification of the commentary https://github.com/OCA/sale-reporting/pull/289#issuecomment-2434451323, I see it fine. Nice Job 👍

rafaelbn (approved):
/ocabot merge major

PR #291, opened 2024-11-06
[FIX] sale_layout_category_hide_detail: Improve interoperability, hide discounted price too

Sibert-Aerts:
Installing this module alongside a custom module of our own we noticed it failed.
The cause of this was a few instances of the following:
```xml
<attribute
    name="t-if"
    add="not current_section or current_section.show_line_amount"
    separator=" and "
 />
```
Though these show foresight they also lack parentheses around the epxression, if applied to an element that already has a `t-if` this forms an illogical expression with erroneous results.
e.g. `hide_discount and not current_section or current_section.show_line_amount`
causes an Exception when `hide_discount` is False and `current_section` is None, I hope you can see why.

Additionally, while fixing this I noticed there was a unit price displayed in the sale portal layout when a line is discounted, that wasn't being hidden when it should have been, an extra rule was added for that case.

### DIFF
--- a/sale_layout_category_hide_detail/views/invoice_report_templates.xml
+++ b/sale_layout_category_hide_detail/views/invoice_report_templates.xml
@@ -66,14 +66,14 @@
     <xpath expr="//span[@t-field='line.price_unit']" position="attributes">
         <attribute
                 name="t-if"
-                add="not current_section or current_section.show_line_amount"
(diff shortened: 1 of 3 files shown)


rousseldenis (approved):
LGTM

pedrobaeza (approved):
It seems OK for that interoperability, and OK also the extra price field.

/ocabot merge minor

PR #293, opened 2024-11-27
[15.0][ADD] sale_report_salesperson_from_partner: New module

pilarvargas-tecnativa:
cc @Tecnativa TT51908

@pedrobaeza Can you check if the nomenclatures are correct? Thanks!


sale_report_salesperson_from_partner/report/sale_report.py
```diff
@@ -0,0 +1,20 @@
+from odoo import fields, models
+
+
+class SaleReport(models.Model):
+    _inherit = "sale.report"
+
+    commercial_user_id = fields.Many2one(
```
pedrobaeza:
```suggestion
    user_id_from_partner = fields.Many2one(
```

I discarded `partner_user_id`, as it can be interpreted as "the partner of the user".

pilarvargas-tecnativa: How about user_from_partner_id?

pedrobaeza: Yes, better to keep conventions, thanks.

pedrobaeza (approved):
Tested on runboat + code review

sergio-teruel (changes requested):
Where is the salesperson partner filter?
And the translation?

sale_report_salesperson_from_partner/views/sale_report_views.xml
```diff
@@ -0,0 +1,16 @@
...
+        <field name="inherit_id" ref="sale.view_order_product_search" />
+        <field name="model">sale.report</field>
+        <field name="arch" type="xml">
+            <filter name="User" position="before">
+                <filter
+                    string="Salesperson from Partner"
+                    name="user_from_partner_id"
+                    context="{'group_by':'user_from_partner_id'}"
+                />
+            </filter>
```
carlosdauden, line 16: You can add `<field name=‘user_from_partner_id’` to facilitate the search by this field.

pilarvargas-tecnativa:
> Where is the salesperson partner filter? And the translation?

I had left the translations to the end because of the modifications, especially the nomenclature. The complete flow is now as it is done with the commercial order. Please have a look at it. Thanks!

sale_report_salesperson_from_partner/views/sale_report_views.xml
pilarvargas-tecnativa: Added, thanks!

sale_report_salesperson_from_partner/views/sale_report_views.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="view_order_product_search" model="ir.ui.view">
+        <field name="inherit_id" ref="sale.view_order_product_search" />
+        <field name="model">sale.report</field>
+        <field name="arch" type="xml">
+            <field name="user_id" position="before">
```
sergio-teruel: Please, put this field after.

sergio-teruel (changes requested):

sale_report_salesperson_from_partner/views/sale_report_views.xml
pilarvargas-tecnativa: Done!

carlosdauden (approved):

sergio-teruel (approved):

PR #297, opened 2024-12-04
[15.0][IMP] sale_report_delivered: Allow salesman to have access to the sale delivery report in the same way as they have access to the sales report.

pilarvargas-tecnativa:
cc @Tecnativa TT51606

@sergio-teruel @carlosdauden please review

### DIFF
--- a/sale_report_delivered/security/ir.model.access.csv
+++ b/sale_report_delivered/security/ir.model.access.csv
@@ -1,2 +1,3 @@
 id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
 access_sale_report_delivered_manager,sale.report.delivered,model_sale_report_delivered,sales_team.group_sale_manager,1,1,1,1
+access_sale_report_delivered_salesman,sale.report.delivered,model_sale_report_delivered,sales_team.group_sale_salesman,1,0,0,0
--- a/sale_report_delivered/__manifest__.py
+++ b/sale_report_delivered/__manifest__.py
@@ -11,5 +11,9 @@
     "installable": True,
     "development_status": "Beta",
     "maintainers": ["sergio-teruel"],
-    "data": ["security/ir.model.access.csv", "views/sale_report_delivered_views.xml"],
+    "data": [
+        "security/ir.model.access.csv",
+        "security/sale_report_security.xml",
+        "views/sale_report_delivered_views.xml",
+    ],
 }
--- a/sale_report_delivered/security/sale_report_security.xml
+++ b/sale_report_delivered/security/sale_report_security.xml
@@ -1,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="sale_report_delivered_personal_rule" model="ir.rule">
+        <field name="name">Personal Delivered Analysis</field>
+        <field name="model_id" ref="model_sale_report_delivered" />
+        <field
+            name="domain_force"
+        >['|',('user_id','=',user.id),('user_id','=',False)]</field>
+        <field name="groups" eval="[(4, ref('sales_team.group_sale_salesman'))]" />
+    </record>
+
+    <record id="sale_report_delivered_see_all_rule" model="ir.rule">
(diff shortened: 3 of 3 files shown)


pedrobaeza:
Didn't you need specific record rules?

pilarvargas-tecnativa:
> Didn't you need specific record rules?

Yes, I was already forgetting... Thanks

pedrobaeza:
The `1 = 1` rule for managers is still missing.

pilarvargas-tecnativa:
> The `1 = 1` rule for managers is still missing.

Done! The tests were already saying that this rule was missing.

carlosdauden (approved):

sergio-teruel (approved):

sale_report_delivered/security/ir.model.access.csv
```diff
@@ -1,2 +1,3 @@
 id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
 access_sale_report_delivered_manager,sale.report.delivered,model_sale_report_delivered,sales_team.group_sale_manager,1,1,1,1
```
pedrobaeza, line 2: I'm thinking that you don't need this ACL line, and in fact, you should never grant delete permissions to `_auto = False` models.

carlosdauden: https://github.com/odoo/odoo/blob/32906d3ef32db9eb7e03e7ac768e3900e89b9c2c/addons/sale/security/ir.model.access.csv#L21-L22

pedrobaeza: Look at this:

https://github.com/odoo/odoo/blob/5a6095cf332bf385092f16b2a11c553cd48767b4/addons/account/security/ir.model.access.csv#L7-L9

Seeing this one, you don't need as well modify not create permissions, just reading ones.

pedrobaeza: That's 18.0 though. On 15.0, it' similar to yours (but only with reading):

https://github.com/odoo/odoo/blob/3a28e5b0adbb36bdb1155a6854cdfbe4e7f9b187/addons/account/security/ir.model.access.csv#L18-L20

I would bet that this is not a change in the framework, but a cleaning...

Is it worth trying? You tell me. I don't want to block this.

carlosdauden: In 18.0, we should look at this:
https://github.com/odoo/odoo/blob/18.0/addons/sale/security/ir.model.access.csv#L49

We normally recommend developers to do the following:
 - If this is the case in 15.0, it is always advisable to ‘mimic’ the version.
 - If it is wrong or can be improved, we can always do a PR to Odoo to change it and then adapt our modules.

If set as is in 18.0 with only this record:
`access_sale_report_salesman,sale.report,model_sale_report,sales_team.group_sale_sale_salesman,1,0,0,0,0,0`
should work without problems and with one less line, but I understand that we should not make people too confused so that they don't lose the spirit of contribution.

@pilarvargas-tecnativa , if you consider it appropriate, make the change, otherwise it's fine.

pilarvargas-tecnativa: I think it's good to follow the Odoo line but leave the read-only permission for the salesman group. Pushed change

pedrobaeza: > If it is wrong or can be improved, we can always do a PR to Odoo to change it and then adapt our modules.

That's the question. It's not wrong. It's just that it seems not needed. Odoo is not going to accept that "improvement"/cleaning on stable versions, so if we detect that something can be done, the sooner, the better. If not, when migrating this module to 18, any contributor (being us or any other) won't probably do the change.

> but I understand that we should not make people too confused so that they don't lose the spirit of contribution.

This is not a random contributor. This is our team, and thus, we can set higher quality standards between us.

pilarvargas-tecnativa: Done in v18
https://github.com/OCA/sale-reporting/pull/314

pedrobaeza (approved):
/ocabot merge minor

PR #305, opened 2025-01-17
[18.0][MIG] sale_report_country_state: Migration to 18.0 

BhaveshHeliconia:


victoralmau (changes requested):
This module is not necessary, as the field https://github.com/odoo/odoo/blob/74eecadb39e0c260f4cfc1ae9478c84eda815132/addons/sale/report/sale_report.py#L48C5-L48C13 already exists in core. 
Did you check it before starting the migration?

BhaveshHeliconia:
Yes, you’re right—this module is no longer needed.

PR #309, opened 2025-01-17
[17.0][MIG] sale_report_country_state: Migration to 17.0

BhaveshHeliconia:


victoralmau (changes requested):
This module is not necessary, as the field https://github.com/odoo/odoo/blob/f66c7c55be5d6db52effafbde7947fb0d6855b79/addons/sale/report/sale_report.py#L50 already exists in core.
Did you check it before starting the migration?

BhaveshHeliconia:
Yes, you’re right—this module is no longer needed.

PR #324, opened 2025-03-20
[FIX] [16.0] sale_report_delivered: Exclude done moves with 0.0 quantities done

Shide:
Done moves are splitted in 2 when quantity done exceedes quantity reserved (but on the interface only shows one). 
When valuation layer is confirmed, only is linked to the first move, using the quantity done and not the product_uom_qty.
So, if these qty done 0.0 moves are excluded, the one with all quantity done will be propagated across the report.

MT-9336 @moduon @rafaelbn @yajo @fcvalgar @Gelojr @sergio-teruel please review if you want 😄

### DIFF
--- a/sale_report_delivered/reports/sale_report.py
+++ b/sale_report_delivered/reports/sale_report.py
@@ -153,13 +153,13 @@ class SaleReportDeliverd(models.Model):
             END AS signed_qty,
             (CASE
                 WHEN t.type IN ('product', 'consu')
-                THEN COALESCE(-svl.quantity, sm.product_uom_qty, 0.0)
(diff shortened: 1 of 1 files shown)


sergio-teruel (approved):

Shide:
@sergio-teruel I've used ABS to convert sub_select values to absolute because in select query will be mutliplied by the sign.

sergio-teruel (approved):

fcvalgar (approved):
Functional review.

LGTM, thank you @Shide

rafaelbn (approved):
/ocabot merge patch

PR #329, opened 2025-06-12
[16.0][FIX] sale_layout_category_hide_detail: compatiblity with `sale_order_line_no_print`

yajo:
The new `sale_order_line_no_print` also searches for the same nodes. Since this module was altering the `if` definition in the same views, you couldn't install both modules at the same time.

This way of overriding the views produces the same result, but doesn't alter the `if` definition, so it is compatible with the other modules.

@moduon MT-7780

### DIFF
--- a/sale_layout_category_hide_detail/views/sale_order_report_templates.xml
+++ b/sale_layout_category_hide_detail/views/sale_order_report_templates.xml
@@ -5,42 +5,31 @@
   <template
         id="report_saleorder_document_hide_detail"
         inherit_id="sale.report_saleorder_document"
+        priority="99"
     >
     <xpath
             expr="//tbody[hasclass('sale_tbody')]//t[@t-if='not line.display_type']"
-            position="attributes"
+            position="replace"
         >
-      <attribute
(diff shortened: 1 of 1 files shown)


pedrobaeza (changes requested):
This is a breaking change and will break any inheritance over the views, and using `replace` is always a bad idea. Please modify instead `sale_order_line_no_print` to fit to your needs and be compatible instead of modifying this one. You can hook onto `get_views` method for such goal.

yajo:
Using replace for wrapping should not break as much as using attributes for
changes in this particular case, because the if is the way to find the
node, wrapped or not.

 inheriting get views for such a small task seems like a too complicated
solution, when this one works fine.

Could you reconsider this fix please? Or do you have any other idea that is
simpler?

pedrobaeza:
As said, overriding `get_views` method in the other module and injecting what's needed can be the option, but this module shouldn't be touched (no matter if the replace adds as wrapper), as it's used widely and it has been here for a long time, as for sure this will break extensions over it.

PR #330, opened 2025-06-24
[18.0][MIG] sale_report_delivered_subtotal: Migration to version 18.0

pilarvargas-tecnativa:
@Tecnativa TT54305

@carlos-lopez-tecnativa @juancarlosonate-tecnativa please review


carlos-lopez-tecnativa (approved):

sale_report_delivered_subtotal/reports/sale_report.py
```diff
@@ -16,17 +16,13 @@ def _select_additional_fields(self):
             sum(
                 CASE WHEN l.price_subtotal <> 0
                     THEN (l.price_subtotal / l.product_uom_qty)
-                    ELSE l.price_reduce
+                    ELSE l.price_unit * (1.0 - l.discount / 100.0)
```
juancarlosonate-tecnativa: Non-blocking comment, i believe you could use `price_reduce_taxexcl` here

juancarlosonate-tecnativa (approved):

sale_report_delivered_subtotal/reports/sale_report.py
pilarvargas-tecnativa: I'm not sure if the result will be the same, for example if the price includes taxes. It may change the result of what was previously obtained.

juancarlosonate-tecnativa: But the price_unit doesn't include taxes, right?

carlos-lopez-tecnativa: According to the code https://github.com/odoo/odoo/blob/d83801027c9a88b6e4db8166b6e2aad41ed760b9/addons/sale/models/sale_order_line.py#L798-L805, there are the fields `price_reduce_taxinc` and `price_reduce_taxexcl`, which refer to the reduced price with and without taxes, respectively. These fields are equivalent to the ones used in your code, so it makes sense to use the ones proposed by @juancarlosonate-tecnativa.

pilarvargas-tecnativa: thanks!

marcelsavegnago (approved):

PR #332, opened 2025-07-08
[FIX] sale_layout_category_hide_detail: portal translation error

jsanchezBinovo:
Fix the xpath so that it can find the item when translating the sales order view from the portal.


sale_layout_category_hide_detail/__manifest__.py
```diff
@@ -5,7 +5,7 @@
     "name": "Sale layout category hide detail",
     "summary": "Hide details for sections in sale orders and invoices for "
     "reports and customer portal",
-    "version": "16.0.1.1.2",
+    "version": "16.0.1.2.2",
```
DantePereyra, line 8: I believe the manifest should not be modified manually, as it is automatically handled by the OCA GitBot.

DantePereyra (changes requested):
Could you please provide more context on why this fix is necessary? A screenshot would be helpful. 
Add branch target (16.0 in this case) on PR tittle to be more easy to find and identify
Also, please review the commit history carefully — you should not include the merge commit. Only the actual fix commit is relevant. 
Check [CONTRIBUTING PAGE](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#git) to see more of this suggestions.

rousseldenis:
@jsanchezBinovo

PR #334, opened 2025-07-28
[18.0][ADD] sale_order_report_without_tax_col

DantePereyra:
Hides taxes data in sale order report if field is unchecked.
- Tax field checked
imagen

-  Tax field unchecked
imagen


chienandalu (commented):
A nice to have would be a context (CONTEXT.md file for the readme) on why this is useful. In our case (this is why I came to your PR) is a customer which quotation taxes are always the same (and normally 0% as they are EU intracomunitary operations), and for them having the summary is enough).

In the other hand, I'm not sure if it's very legal to hide the taxes summary...

My take on this (if your context is similar) would be to have a computed field in the sale.order for a boolean flag that is on when all the taxes are the same and thus they would be redundant in the order line details.

cc @rafaelbn (MT-10378)

rafaelbn:
I would like to think about the naming of the module. Thank you for this great contribution @DantePereyra ! ❤️

Gelojr (changes requested):
Requesting changes: I ask for a review because with the “Taxes” option enabled the sale order report does not display the per-line taxes column, while the tax summary is shown; the standard Odoo (runbot) report displays the taxes column, so the module diverges from expected behavior.
image

The following tests were executed:

- Test 1 – OK: The taxes visibility toggle is available under Settings > Sales > Taxes and can be turned on/off.
- Test 2 – OK: With “Taxes” disabled, the sale order report hides both the per-line taxes column and the tax summary.
- Test 3 – NOT OK: With “Taxes” enabled, the sale order report still hides the per-line taxes column while the tax summary is shown; the standard Odoo (runbot) report displays the taxes column.
- Test 4 – OK: The setting persists across sessions/users and applies to the report.
- Test 5 – NOT OK: On an order with mixed taxes, with “Taxes” enabled the per-line taxes column remains hidden while the tax summary is shown; in the standard Odoo (runbot) report the taxes column is displayed.
- Test 6 – OK: A user with configuration rights can change the setting without access errors.
- Test 7 – OK with note: Spanish translations are present for the setting and main report texts; generic strings like “Total” come from Odoo’s base translation, not this module.

DantePereyra:
Hi @Gelojr and @rafaelbn, thanks for your time!
I’ve just left a [comment](https://github.com/OCA/sale-reporting/pull/337#issuecomment-3210032380) that might be of interest to you—would appreciate your feedback there.

rafaelbn:
Hi @DantePereyra 😄 !

I answered in https://github.com/OCA/sale-reporting/pull/337#issuecomment-3213321432

🫶

Issue #333, opened 2025-07-09
[18.0][sale_order_report_product_image] images are not displayed in the PDF.


achulii:
## Module
sale_order_report_product_image

## Describe the bug
When adding an image to a product, it does not appear in the generated Sales Order (SO) PDF report, but it does show up correctly in the HTML version. The issue does not occur with demo products — their images are displayed as expected.

## To Reproduce
**Affected versions**: 18.0

1. Add an image to a custom product.
2. Generate the Sales Order report (PDF).

**Additional context**
Tested on runboat.

BT-jmichaud:
I can confirm the issue exists. When printing a product with image, there is only an empty square.

odooexperts-consultants:
i second that

Rad0van:
@pedrobaeza sorry to bother but I consider you the greatest expert. Do you have any idea what might be going on here? Thanks.

pedrobaeza:
Sorry, I don't use this module.

Rad0van:
> Sorry, I don't use this module.

@pedrobaeza the question was more general. The funny thing is that _image_128_ field is correctly rendered in HTML version of report for both demo products and the ones you add. But in PDF version it only works for demo products while the ones you add render empty box... There's some strange thing going on but I cannot find out what and I thought you maybe could have some idea...

pedrobaeza:
Try using `t-esc` instead of `t-field`.

BT-jmichaud:
PR #335 created

Rad0van:
@pedrobaeza is my hero! Thanks a lot.

PR #335, opened 2025-07-29
[18.0][FIX] sale_order_report_product_image: show the image in PDF

BT-jmichaud:
Fix the problem reported in Issue #333

### DIFF
--- a/sale_order_report_product_image/views/report_saleorder.xml
+++ b/sale_order_report_product_image/views/report_saleorder.xml
@@ -10,7 +10,7 @@
         <td name="td_name" position="before">
             <td name="td_image">
                 <span
-                    t-field="line.product_id.image_128"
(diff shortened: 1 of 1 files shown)


Rad0van (approved):
Works!

BT-vseydoux (approved):
Fix tested and approved

Rad0van:
@BT-jmichaud it still is not pefect :-( In case there is no picture assigned to the product it crashes like this:
`TypeError: argument should be a bytes-like object or ASCII string, not 'bool'`
So you need to add something like
`t-if="line.product_id.image_128"`
into the span as well.

BT-jmichaud:
@Rad0van 
I made a new PR: https://github.com/OCA/sale-reporting/pull/336

PR #337, opened 2025-08-20
[16.0][ADD] sale_order_line_hide_tax_in_report: new module

chienandalu:
Whenever the taxes are in the same group for the whole order the column won't be displayed.

**Context**

When the taxes are the same in every sales line, that information might be redundant with the taxes summary and takes valuable space that other columns might take.

This might be event more annoying when there's tax exemptions like in EU intracomunitary operations.

**Usage**

Whenever the taxes are in the same group for the whole order the column won't be displayed.

cc @moduon MT-11205

please review @Gelojr @rafaelbn 

fyi @DantePereyra this would be my implemantion commented in https://github.com/OCA/sale-reporting/pull/334#pullrequestreview-3115145369

**TODO**

- [ ] Add tests


Gelojr (approved):
Great job @chienandalu congratulations on the contribution!
The following tests were performed:

- Test 1: Hide taxes with common group—Taxes column is not shown when all product lines share the same tax group on quotation/sales order report.
- Test 2: Show taxes with different groups—Taxes column is shown when at least two lines use different tax groups.
- Test 3: Display lines don’t affect—Sections/notes are ignored; with two product lines in the same group the Taxes column stays hidden.
- Test 4: Multiple taxes in same group—A line with several taxes all in the same group is treated as homogeneous; Taxes column hidden.
- Test 5: Mixed groups within a line—A single line combining taxes from different groups triggers mixture; Taxes column shown.
- Test 6: No product lines (edge case)—With only display lines or an empty order the report keeps the Taxes column visible (no product taxes listed).
- Test 7: After confirmation—Behavior remains identical after confirming quotations; homogeneous orders hide the column and mixed orders show it.

DantePereyra:
Hi! Thanks for sharing your thoughts. I completely see the advantages of your approach.
Would you be open to merging both addons? (#334)

After reviewing the feedback, I was considering adding more personalization options. Specifically:

- Two checkboxes in `res.config.settings`:
  - One to hide the tax column.
  - Another to hide the tax summary.
- These settings would propagate to `res.partner`, where the same two checkboxes would be available.

This way, you could configure it either at company level, at partner level, or both—especially useful for international operations where legal requirements may differ.

What do you think?

chienandalu:
Ok @DantePereyra could you make a PR to this one?

rafaelbn:
@DantePereyra  thanks a lot for sharing ❤️  your perspective and for pushing this forward. I completely understand that sometimes customers request specific configuration options, but as product maintainers we have to evaluate carefully how these requests affect the product design in the long term.

As Julie Zhuo says: “Users should not face dozens of checkboxes in a single form, because that degrades the experience. Configuration must live in the right place, not scattered everywhere.” And in Jason Fried’s words: “Let’s not complicate what can be solved with a single elegant option. If you need more flexibility, look for better rules, not more checkboxes.”

From a **product design perspective**, adding checkboxes in `res.partner` for controlling how sales orders are printed is not the right solution:
- `res.partner` is already overloaded with checkboxes from different modules, and this creates noise and poor UX.  
- The decision of whether to show or hide tax columns in a corporate document is not a *per-partner preference*. It belongs at a higher configuration level (company, fiscal position, industry, or specific tags).  
- Having the possibility to issue two different quotations to the same customer (one with taxes column, one without) is not a realistic scenario in most companies worldwide.  

I believe the objective should be clear: migrations and improvements must serve the product point of view, not just the PR point of view. The goal is consistency, simplicity, and coherence.

So my proposal would be:  
1. Keep the addon simple, allowing to hide the tax column (this is useful and valid).  
2. Avoid introducing partner-level checkboxes. If more flexibility is truly needed, it should be solved through a higher-level configuration such as fiscal position, tags, or company settings — not at `res.partner`.  
3. Ensure we do not add complexity without a validated real-world use case.  

This way we protect both the user experience and the maintainability of the OCA ecosystem.

What do you think? 🤔💙😊

rafaelbn:
I would recommend change the name to `sale_order_report_hide_tax` . What do you think? 🤔

rafaelbn (approved):
- [ ] I would recommend change the name to `sale_order_report_hide_tax` . What do you think? 🤔

Tested in runboat 👍🏼 

imagen

fcvalgar (approved):
Congratulations on the work @chienandalu, the implementation is clean and behaves as intended; the following tests were executed and all passed.

Test 1: OK – Hides the Taxes column when all real lines share the same tax group.
image

image
Test 2: OK – Shows the Taxes column when lines contain different tax groups (e.g., exempt vs VAT).
image

image

Test 3: OK – Sections/notes (display_type) are ignored in the decision.
Test 4: OK – Single real line hides the Taxes column.
Test 5: OK – Quote with only sections/notes renders without a taxes column on lines.
Test 6: OK – Different tax rates within the same group still hide the Taxes column (comparison by tax_group).
Test 7: OK – Product bundle that expands to components with mixed tax groups shows the Taxes column.
Test 8: OK – Fiscal position mapping to a tax in the same group keeps the Taxes column hidden.
Test 9: OK – After confirming the quotation, the Sales Order report preserves the same behavior.
Test 10: OK – QWeb template guards (t-if on th_taxes and td_taxes) are present and functional.

chienandalu:
>  I would recommend change the name to sale_order_report_hide_tax . What do you think? 🤔

@rafaelbn Done

rafaelbn (approved):
Great!

/ocabot merge patch

PR #350, opened 2025-12-22
[18.0][MIG] sale_order_report_hide_tax: migration to 18.0

sabrinaRMartin:
Migration to 18.0

@moduon @chienandalu MT-13132

https://www.loom.com/share/d1e1d20059d544648aeb7f46a594b9ea


chienandalu (changes requested):
Watchout your commit structure:

Image

sale_order_report_hide_tax/tests/test_show_tax_column.py
```diff
@@ -0,0 +1,125 @@
+from odoo.tests.common import TransactionCase
```
chienandalu:
```suggestion
# Copyright 2025 Moduon Team S.L.
# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl-3.0)
from odoo.tests.common import TransactionCase
```

sale_order_report_hide_tax/tests/test_show_tax_column.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+class TestShowTaxColumnInReport(TransactionCase):
```
chienandalu: Use `BaseCase` by default if you don't need tracking
```suggestion
class TestShowTaxColumnInReport(BaseCase):
```

sale_order_report_hide_tax/tests/test_show_tax_column.py
```diff
@@ -0,0 +1,125 @@
...
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+
```
chienandalu: Don't leave blank lines after every statement...
```suggestion
```

sale_order_report_hide_tax/tests/test_show_tax_column.py
```diff
@@ -0,0 +1,125 @@
...
+                            "product_uom_qty": 1,
+                            "price_unit": 50,
+                            "tax_id": [(6, 0, [self.tax_1.id])],
+                        },
+                    ),
+                ],
+            }
+        )
+
+        order._compute_show_tax_column_in_report()
```
chienandalu: You shouldn't need to trigger it manually

sale_order_report_hide_tax/tests/test_show_tax_column.py
```diff
@@ -0,0 +1,125 @@
...
+            {
+                "name": "Tax 2",
+                "amount": 5,
+                "amount_type": "percent",
+                "tax_group_id": cls.tax_group_2.id,
+            }
+        )
+
+    def test_same_tax_group_hides_tax_column(self):
+        order = self.env["sale.order"].create(
```
chienandalu: You can just declare it in the setup and simply change one line taxes

sale_order_report_hide_tax/tests/test_show_tax_column.py
```diff
@@ -0,0 +1,125 @@
...
+                        },
+                    ),
+                ],
+            }
+        )
+
+        order._compute_show_tax_column_in_report()
+
+        self.assertFalse(
+            order.show_tax_column_in_report,
```
chienandalu: What you should really test is the rendered report

sale_order_report_hide_tax/tests/test_show_tax_column.py
```diff
@@ -0,0 +1,125 @@
...
+                        0,
+                        0,
+                        {
+                            "product_id": self.product.id,
+                            "name": "Line 1",
+                            "product_uom_qty": 1,
+                            "price_unit": 100,
+                            "tax_id": [(6, 0, [self.tax_1.id])],
+                        },
+                    ),
```
chienandalu: Better use `Comand.create`

chienandalu (changes requested):
Thanks for the tests. Some changes requested to improve them:

sale_order_report_hide_tax/tests/test_show_tax_column.py
sabrinaRMartin: Should this be `BaseCommon`?

chienandalu: yes, sorry, that's what I meant

Gelojr (changes requested):
The pull request **is not approved**. A review is requested.

**Functional tests performed**

The following functional tests have been executed and their results are correct:
Test 1: Hide tax column when all order lines share the same tax group — OK.
Test 2: Show tax column when different tax groups are used across order lines — OK.
Test 3: Ignore section and note lines when deciding tax column visibility — OK.
Test 4: Multiple taxes in the same tax group on a single line correctly hide the tax column — OK.
Test 6: Single order line with a single tax group behaves correctly — OK.

**Test not passed**

Test 5: Spanish localization with fiscal position adding withholding tax — NOT OK.

**Steps to reproduce the issue**

1. Install and configure the Spanish localization.
2. Use a Fiscal Position that applies a withholding tax (retención) in SO.
3. Create a Sales Order for a customer and assign this Fiscal Position to the order.
4. Add a product line with VAT 21% as the product tax.
5. Verify that Odoo automatically adds the corresponding withholding tax to the sales order line due to the Fiscal Position.
6. Print or preview the Sales Order / Quotation report.

**Observed behavior**

The tax column is not printed in the report, and taxes are not displayed, even though the order line contains taxes belonging to different tax groups (VAT and withholding).

**Expected behavior**

The tax column should be displayed in the report, since the sales order line includes taxes from different tax groups.

Based on the above, the pull request is not approved and a revision is requested to address this scenario.

chienandalu:
> Test 5: Spanish localization with fiscal position adding withholding tax — NOT OK.

@sabrinaRMartin please make a test for this and debug the reason why. At first glance, I'd say it should work, but it isn't.

sabrinaRMartin:
> > Test 5: Spanish localization with fiscal position adding withholding tax — NOT OK.
> 
> @sabrinaRMartin please make a test for this and debug the reason why. At first glance, I'd say it should work, but it isn't.

@chienandalu The problem arose because we were comparing one recordset with another, and it always returned the same values. In other words, for a line that has group A, B, the tax_group_id also returns group A, B, so the condition for displaying the column was not being met because first_line_tax_group and tax_group_id had the same value.

sabrinaRMartin:
@Gelojr @chienandalu could you please check the PR again? Everything should be working fine now.

sale_order_report_hide_tax/tests/test_show_tax_column.py
```diff
@@ -0,0 +1,152 @@
...
+        )
+        html = result[0].decode("utf-8")
+        self.assertIn(
+            "Taxes",
+            html,
+            """The tax column should be displayed when taxes
+                belong to different tax groups""",
+        )
+
+        def test_mixed_tax_groups_single_line_sow_tax_column(self):
```
chienandalu: you've put this method inside the other one. This test won't be executed

sale_order_report_hide_tax/tests/test_show_tax_column.py
```diff
@@ -0,0 +1,152 @@
...
+                            "product_id": cls.product.id,
+                            "name": "Line 1",
+                            "product_uom_qty": 1,
+                            "price_unit": 100,
+                            "tax_id": [(6, 0, [cls.tax_1.id, cls.tax_2.id])],
+                        },
+                    ),
+                ],
+            }
+        )
```
chienandalu: You don't need to create that many orders. Just create one an play with its lines...

sale_order_report_hide_tax/models/sale_order.py
```diff
@@ -15,8 +15,15 @@ def _compute_show_tax_column_in_report(self):
...
-            order.show_tax_column_in_report = (
-                first_line_tax_group != order_lines.tax_id.tax_group_id
+            first_line_tax_group = fields.first(order_lines).tax_id.mapped(
+                "tax_group_id"
             )
+            # Check if any order line requires displaying the tax column
+            for line in order_lines:
+                line_tax_groups = line.tax_id.mapped("tax_group_id")
+                if len(line_tax_groups) > 1:
+                    break
```
chienandalu: Why?

sale_order_report_hide_tax/models/sale_order.py
```diff
@@ -15,8 +15,15 @@ def _compute_show_tax_column_in_report(self):
...
+                if line_tax_groups != first_line_tax_group:
+                    break
+            else:
+                order.show_tax_column_in_report = False
```
chienandalu: `for..else` is quite a confusing construc. Even python's father despise it 😅 : https://mail.python.org/pipermail/python-ideas/2009-October/006157.html

chienandalu: Simple fix to the previous code
```suggestion
            first_line_tax_group = fields.first(order_lines).tax_id.tax_group_id
            # Mixed group taxes, let's show them for clarity
            order.show_tax_column_in_report = any(
                first_line_tax_group != line.tax_id.tax_group_id for line in order_lines
            )

File: sale_order_report_hide_tax/tests/test_show_tax_column.py [hunk 12]
```diff
@@ -0,0 +1,152 @@
...
+        def test_mixed_tax_groups_single_line_sow_tax_column(self):
+            result = self.report._render_qweb_html(
+                "sale.report_saleorder", [self.order_3.id]
+            )
+            html = result[0].decode("utf-8")
+            self.assertIn(
+                "Taxes",
+                html,
+                """The tax column should be displayed when a single
+                order line has taxes belonging to different tax groups""",
```
chienandalu: Why? That's not a premise of the feature

chienandalu (changes requested):

sale_order_report_hide_tax/tests/test_show_tax_column.py
sabrinaRMartin: > Test 5: Spanish localization with fiscal position adding withholding tax — NOT OK.

We are testing that when there is a single order line with taxes from different tax groups, these are displayed.

chienandalu: That's not the issue. Read carefully the STR. 

> 1. Install and configure the Spanish localization.
> 2. Use a Fiscal Position that applies a withholding tax (retención) in SO.
> 3. Create a Sales Order for a customer and assign this Fiscal Position to the order.
> 4. Add a product line with VAT 21% as the product tax.
> 5. Verify that Odoo automatically adds the corresponding withholding tax to the sales order line due to the Fiscal Position.
> 6. Print or preview the Sales Order / Quotation report.
>
> Expected behavior: The tax column should be displayed in the report, since the sales order line includes taxes from different tax groups.

sabrinaRMartin: > Expected behavior: The tax column should be displayed in the report, since the sales order line includes taxes from different tax groups.

@chienandalu  We want to verify that when we have a single product with more than one tax in different tax groups, the column appears. Why is it not correct to implement this test with this logic?

Gelojr (changes requested):
After re-running the test where two taxes from different tax groups are applied to the sales order line due to the fiscal position, the taxes column is still not printed in the report despite the taxes belonging to different tax groups.


Image

Image

Image

sabrinaRMartin:
@Gelojr when we have a single product line with taxes from different groups, so the column should not be displayed. 
Since a tax position is defined in the sales order, all products are subject to the tax defined by the tax position. In any case, the column should be displayed if the tax on the products belongs to different tax groups.

For example:
A. The column is displayed:
1. You configure two taxes in different groups (Group A and Group B).
2. You create a quote with two product lines that have a tax (one from each group).
3. You define a tax position for the sales order (or directly for the customer).

Here the column should be displayed.

B. The column is NOT displayed:
1. You create a quote with 2 product lines that have a tax (from the same group).
2. You define a tax position for the sales order (or directly on the client).

Here, the column should NOT be displayed.

I will modify the README to make it clear how it works.

Gelojr (approved):
Great work on this contribution @sabrinaRMartin. The functional behavior is well thought out and aligned with the goal of avoiding redundant information in the report.

The following tests have been executed to validate the module behavior after the latest changes:

- Test 1: Hide tax column when all order lines use taxes from the same tax group — OK
- Test 2: Show tax column when different tax groups are used across order lines — OK
- Test 3: Ignore section and note lines when evaluating tax groups — OK
- Test 4: When two taxes from different tax groups are applied on one line or across multiple lines, the tax column is still hidden as long as the tax groups are the same for the whole lines of the sale order (to avoid duplication with the tax summary). Example: Spanish localization fiscal position adds withholding tax (retención) consistently together with VAT — OK
- Test 5: Different taxes belonging to the same tax group on different lines correctly hide the tax column — OK

chienandalu (approved):
Thanks @sabrinaRMartin !

Andrii9090-tecnativa (approved):
Code&functional review 👍 

Image

rafaelbn (approved):
Funcional test 👍🏼 

/ocabot merge patch

PR #351, opened 2026-01-13
[19.0][MIG] base_multicompany_reporting_currency: migration to 19.0

SilvioC2C:
Examples of the settings view:

- user w/ multicurrency access

Screenshot from 2026-01-15 12-06-49

- user w/o multicurrency access

Screenshot from 2026-01-13 11-56-08


vvrossem (approved):
LG

base_multicompany_reporting_currency/data/ir_config_parameter.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <record id="multicompany_reporting_currency" model="ir.config_parameter">
+        <field
+            name="key"
+        >base_multicompany_reporting_currency.multicompany_reporting_currency</field>
+        <field name="value" eval="int(ref('base.EUR'))" />
+    </record>
```
ivantodorovich: This is not related to the migration, but..

Won't this parameter be over-written upon a module update, everytime?
Given it doesn't have `noupdate="1"`

What happens if you choose a different currency, and then update the module?

I'm more inclined to have this parameter be created by a `post_init_hook` or an `init` override like this one: https://github.com/odoo/odoo/blob/298c96045ecaff6cdd89cf192d5ed3a7eb931e98/odoo/addons/base/models/ir_config_parameter.py#L44-L57

ivantodorovich (approved):

base_multicompany_reporting_currency/data/ir_config_parameter.xml
SilvioC2C: I didn't realise the `noupdate` is missing. And now that I think of it, one might also delete this param manually from the System Parameters list view, recreate it manually (without the XMLID), and the next time the module is upgraded, Odoo will crash trying to recreate the param with the same key (which must be unique). 


I will add a `init` override as you suggested.

SilvioC2C:
@ivantodorovich @vvrossem can you please update your reviews?

While working on `crm_multicompany_reporting_currency` mig to v19, I found out the architecture of this module was not good enough to make everything work correctly, so I've made some major improvements to migrate it.

vvrossem (approved):
:superhero:

ivantodorovich (approved):

BhaveshHeliconia (approved):
LGTM!

vvrossem:
Hello @rousseldenis , sorry for contacting you directly. Do you think this PR can be merged as well? :pray:

rousseldenis (approved):

PR #354, opened 2026-02-16
[19.0][MIG] sale_layout_category_hide_detail

glitchov:
Migration from 18.0


baimont:
image

image

as discussed with you @glitchov we lose the standard v19 3 points button when adding a section. we'll see how we manage to fix that. thanks for your work!

rousseldenis:
@glitchov @baimont What's the status of this ?

baimont:
> @glitchov @baimont What's the status of this ?

hey @rousseldenis 

not used anymore by the client as a part of the feature (the one to hide prices) is now part of odoo

PR #356, opened 2026-03-10
[18.0][ADD] sale_report_product_tag: Add product tag field in Sale Report

cvinh:
This module extends the **Sales Analysis** report (``sale.report``) to
allow grouping and filtering by **Product Tag**.

Known issues / Roadmap
======================

Since product tags are a Many2many field, we had to choose to take the
first tag in the products. This avoids any double-counting of sales
figures that would occur with a direct Many2many field. This can be
improved if we want real figures for every tag in the products


cvinh:
can you reopen please ?

PR #378, opened 2026-06-12
[FIX] [16.0] sale_report_delivered: Restore Dropshipping and fix calculation on modified done moves

Shide:
Restore Dropshipping calculation.
Fixes the report calculation when incoming/outgoing done moves are reduced. The sign is now adjusted based on the remaining quantity generated by the new inventory valuation.

**Added Known Issues / Roadmap (I think it's an Odoo bug because should increase or decrease remaining_qty)**

https://www.loom.com/share/bb46ecd3868a45caaa5e606636e91e42

MT-14985 @moduon @sergio-teruel @carlosdauden @chienandalu @fcvalgar @Gelojr please review if you want 😄

### DIFF
--- a/sale_report_delivered/tests/test_sale_report_delivered.py
+++ b/sale_report_delivered/tests/test_sale_report_delivered.py
@@ -29,56 +29,275 @@ class TestSaleReportDeliveredBase(common.TransactionCase):
         group_sale_manager = cls.env.ref("sales_team.group_sale_manager")
         group_sale_manager.write({"users": [(4, cls.admin.id)]})
 
-        cls.product = cls.env["product.product"].create(
-            {"name": "Test product", "type": "product", "list_price": 10}
+    @classmethod
+    def _create_product(
+        cls, name, ptype, list_price=10.0, standard_price=0.0, stock_qty=0
+    ):
+        """Create a product and optionally add stock for storable products."""
+        ptype = ptype or "product"
+        product = cls.env["product.product"].create(
+            {
+                "name": name,
+                "type": ptype,
+                "list_price": list_price,
+                "standard_price": standard_price,
+            }
         )
-        cls._create_stock_quant(cls, cls.product)
-        cls.service = cls.env["product.product"].create(
-            {"name": "Test service", "type": "service", "list_price": 10}
-        )
-        cls.order_1 = cls._create_order(cls, cls.product)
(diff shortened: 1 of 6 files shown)


chienandalu (commented):
Could we have the dropship conditions tested as well to avoid future regressions?

Shide:
> Could we have the dropship conditions tested as well to avoid future regressions?

Yes we can! I'm on it

sale_report_delivered/tests/test_sale_report_delivered.py
```diff
@@ -87,7 +306,7 @@ def _test_sale_report_delivered_read_group(self):
             ],
             groupby=["order_id"],
         )
-        self.assertAlmostEqual(res[0]["margin_percent"], 100.00)
+        self.assertAlmostEqual(res[0]["margin_percent"], 70.00)
```
chienandalu, line 309: What changed in the test that is making the validation value changing?

Shide: Don't know why happens (maybe for class methods...)
Also, using standard price covers the use in Odoo. 
No need to reevaluate layers because creates layers correctly.

chienandalu (approved):
Thanks :)

fcvalgar (approved):
Great work, thank you for the fix @Shide .

I reviewed the module focusing on the restored dropshipping behavior and the delivered sales report calculation. The following tests were performed:

Test 1: OK - Dropship sale fully delivered. The delivered sales report correctly displayed the dropship sale with the expected delivered quantity and delivered amount.
Test 2: OK - Partially delivered dropship sale. The report used the actual delivered quantity and did not report the full ordered quantity.
Test 3: OK - Dropship return with refund. The report correctly reflected the positive delivered quantity and the negative returned quantity, showing the expected net delivered quantity.
Test 4: OK - Dropship return without refund. The return did not incorrectly reduce the delivered sales report values.
Test 5: OK - Modified done dropship move quantity. After changing the done quantity on a completed dropship move, the report recalculated the delivered quantity correctly.
Test 6: OK - Dropship delivered margin. The report displayed coherent delivered amount, cost and margin values for the dropship sale.

Everything worked as expected for the tested dropshipping flows, so I approve this PR.

Image

rafaelbn (approved):
👍🏼 

/ocabot merge patch

PR #379, opened 2026-06-17
[18.0][IMP] sale_order_product_recommendation_product_sold_by_delivery_week: show live per-customer weekly sales in catalog

CarlosRoca13:
The product catalog showed the globally stored (and cron-refreshed) weekly sales hint from product_sold_by_delivery_week, which did not match the recommendation wizard: the wizard computes the hint live and filtered by the order's customer.

When the catalog is opened from a sale order it now displays a new field, weekly_sold_delivered_catalog_shown, computed on the fly and filtered by the order's commercial partner, hiding the base global hint. Outside of a sale order catalog the product lists keep showing the original stored field, and the base module is left untouched.

The computation shared by the wizard and the catalog is extracted into a single helper, product.product._weekly_sold_delivered_shown_map(), so the hint is produced in one place and both computes just map their records.

cc @Tecnativa TT63163

ping @sergio-teruel @carlosdauden

### DIFF
--- a/sale_order_product_recommendation_product_sold_by_delivery_week/models/product_product.py
+++ b/sale_order_product_recommendation_product_sold_by_delivery_week/models/product_product.py
@@ -1,0 +1,52 @@
+# Copyright 2026 Tecnativa - Carlos Roca
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import api, fields, models
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    weekly_sold_delivered_catalog_shown = fields.Char(
+        string="Weekly Sold for Catalog",
+        compute="_compute_weekly_sold_delivered_catalog_shown",
+        groups="sales_team.group_sale_salesman",
+    )
+
+    def _weekly_sold_delivered_shown_map(self, partner):
+        """Return ``{product: formatted weekly string}`` for ``self`` computed
+        live and filtered by ``partner`` (its commercial partner). Services are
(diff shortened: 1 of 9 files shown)


sale_order_product_recommendation_product_sold_by_delivery_week/models/product_product.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class ProductProduct(models.Model):
+    _inherit = "product.product"
+
+    weekly_sold_delivered_catalog_shown = fields.Char(
+        string="Weekly Sold for Catalog",
+        compute="_compute_weekly_sold_delivered_catalog_shown",
+        groups="sales_team.group_sale_salesman",
+    )
```
chienandalu, line 13: Doesn't this fit in the base module?

CarlosRoca13: You're right that there's no hard technical dependency: the catalog field doesn't use anything from sale_order_product_recommendation and the base module already has the catalog view and the weekly_partner_id filtering, so it could live there.

I kept it in the glue module on purpose, though. Its only reason to exist is to make the catalog match the recommendation wizard. And the base field is intentionally stored + cron-refreshed for performance, showing a global hint; moving this in would force a live, per-customer recompute on every catalog open for everyone using only the base module, changing its behavior and performance. So the field belongs where the "match the wizard" intent lives.

If you'd still prefer the global catalog hint to be live/per-customer for base users too, I can move the field + view there instead.

